PR #6, opened 2016-09-03
allow source to specify candidate kind

bfredl:
I tried to create denite source with separate kind (bfredl/nvim-miniyank#4) but it didn't work. Seems "jump_list" is hard coded. I imagine the intended behavior to be something like this.


Shougo:
Implemented.

PR #15, opened 2016-09-24
Ack source

carlitux:
Same source as grep but using ack

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -290,6 +290,12 @@ filetype	Gather filetypes and change the filetype of the current
 						*denite-source-grep*
 grep		Gather grep results and nominates them.
 
+						*denite-source-ack*
+ack		Gather ack results and nominates them.
+		Examples when prompts for query:
+			* pattern
+			* --[type] pattern [relative path to current path]
+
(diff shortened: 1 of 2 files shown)


zchee:
@carlitux Ah, I don't totally understand denite implements, and not read the your PR.
But maybe can change the `grep` command to other command, like
```vim
call denite#custom#source('grep', 'command', ['pt', '--follow', '--nocolor', '--nogroup', '--hidden', '-g', ''])
```

@Shougo It's correct? denite.nvim will ignore(or not use) it even if I set this config?

Shougo:
@carlitux Why do you add the command source commit?
It is not related Ack source.  Please create new PR for it.

And yes, the ack source is not needed.
You can change the grep command by `denite#custom#var()`.

carlitux:
@Shougo sorry I didn't to correct use of branches

carlitux:
@Shougo @zchee maybe changing the code on grep. It is not working as I need using my ackrc config.

With this pull request I can query like '--js React src' not only 'ack'

Shougo:
I have added the additional arguments support in grep source.
You don't have to create new source.

PR #17, opened 2016-09-25
Grep improves

carlitux:
This allows to use shell environment variables and configs.

Searches under current work directory and with this changes I can use ack:

```vim
call denite#custom#var('grep', 'command', ['ack'])
call denite#custom#var('grep', 'recursive_opts', [])
call denite#custom#var('grep', 'separator', [])
call denite#custom#var('grep', 'final_opts', [])
call denite#custom#var('grep', 'default_opts', 
     \ ['-s', '-H', '--nopager', '--nocolor', '--nogroup', '--column'])
```

```vim
:Denite grep

Pattern: --[ack-type] React [subdirectory or any other extra param ]
```

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -96,6 +96,17 @@ EXAMPLES						*denite-examples*
 		\ ]
 
 	call denite#custom#var('menu', 'menus', s:menus)
+
+	" Ack command on grep source, doesn't need some vars defaults are 
+	" enough
+	call denite#custom#var('grep', 'command', ['ack'])
+	call denite#custom#var('grep', 'recursive_opts', [])
+	call denite#custom#var('grep', 'final_opts', [])
+	call denite#custom#var('grep', 'separator', [])
+	call denite#custom#var('grep', 'default_opts',
+			\ ['--ackrc '.$HOME.'/.ackrc', '-H',
+			\ '--nopager', '--nocolor', '--nogroup', '--column'])
+
 <
 
 ==============================================================================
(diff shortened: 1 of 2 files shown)


Shougo:
Please remove the command source commit from it.
And please check the latest grep source.

Shougo:
I don't want to use shell.  And your changes fails if the arguments contain spaces.

Why this change is needed?
Please describe it.

carlitux:
Well if running without shell I can't use .ackrc... at least not in the test I did.

> I don't want to use shell. And your changes fails if the arguments contain spaces.

Do you have example when will fails?

Shougo:
Why ack can't use .ackrc without shell?
It seems ack problem.

Shougo:
I don't want to consider shell behavior.
If grep source use shell, it depends on shell behavior.
To support Windows is difficult.

carlitux:
@Shougo sound good reasons, please review this change so new commands can use any subdirectory not only root.

I don't know if grep is the only command that need search path by default

Shougo:
Why the `final_opts` needed?
You want to omit the directory argument.  Why?

carlitux:
Grep needs path https://github.com/carlitux/denite.nvim/blob/master/rplugin/python3/denite/source/grep.py#L44

```zsh
grep -inH -r -- pattern /path/to/directory
```

Ack doesn't needs it 

```zsh
ack -- pattern
```

but I can have the flexibility to specify any subdirectory to do quicker searches or avoid large files like min css, js files

```zsh
ack -- pattern subdirectory
```

Shougo:
Please add `final_opts` documentation.
And you should add the settings for ack in the example section.

Shougo:
The grep source is updated.  Please resolve the conflict.

carlitux:
done

Shougo:
Merged.

PR #19, opened 2016-09-25
Improve grep source

carlitux:
- Update doc to enable ackrc in config
- Input parsed by shlex to use all the power of grep-like programs,
  allow productivity when needs to add more arguments to default_opts
  and :Denite grep [arguments].
- Extend timeout to 2 seconds instead 0.1 (no results with large project),
  we can use 2 seconds because now is async.

### DIFF
--- a/rplugin/python3/denite/source/grep.py
+++ b/rplugin/python3/denite/source/grep.py
@@ -8,6 +8,7 @@ from .base import Base
 from denite.util import parse_jump_line
 import subprocess
 import os
+import shlex
 
 
 class Source(Base):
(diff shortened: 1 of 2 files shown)


rplugin/python3/denite/source/grep.py
```diff
@@ -64,7 +65,8 @@ def gather_candidates(self, context):
 
     def __async_gather_candidates(self, context):
         try:
-            outs, errs = self.__proc.communicate(timeout=0.1)
+            outs, errs = self.__proc.communicate(timeout=2)
+            self.debug(errs)
```
Shougo: Please remove the debug message.

Shougo:
I will merge it after you have remove the debug message.

carlitux:
@Shougo here is the change, sorry I forgot do it before commit

Shougo:
LGTM

PR #20, opened 2016-09-26
Async gather

tjdevries:
@Shougo, In the case of multiple different sources being used, this will execute them asynchronously from each other in different threads.

So if you had an IO blocking source, you would not be waiting for that to complete before starting the gather candidate phase for a different source.

I also included a fix to the menu source I submitted, since I found it while I was creating this PR. I can move it to a different one if you want.


Shougo:
Nice feature.
It seems useful for deoplete.

PR #21, opened 2016-09-26
Revert "Async gather"

Shougo:
Reverts Shougo/denite.nvim#20

It fails if you use neovim and the external sources (for example: file_mru and neoyank).


Shougo:
Reverted.

PR #25, opened 2016-10-04
Add directory_rec source

okamos:
## Summary
* Add directory_rec source
* Add directory kind

## Detail
* directory_rec gather directories recursive.
* directory kind execute vim cd command, but unite.vim narrow function is better than this action.
* Update documentation


Shougo:
LGTM

PR #30, opened 2016-10-15
feat(): add help source

mhartington:
Add source for help


Shougo:
I think the source should analyse `doc/tags` files like unite-help.
https://github.com/tsukkee/unite-help

And no documentation for help source.

mhartington:
Will add the docs.
Wasn't sure on the tags part but will look into how to add it

mhartington:
Fixing the conflicts now. 
How does this look for search through the tags file?

PR #31, opened 2016-10-17
feat(): add help source

mhartington:
let's give this another try 😄


Shougo:
I have tested it.  LGTM.

PR #33, opened 2016-10-20
Disable relativenumber to completely remove numbers

rafi:
# Problem
If a window has `relativenumber` enabled, denite's window displays numbers. Screenshot: 
image

# Expected
Denite's window should have numbers disabled, probably custom icons are planned.


Shougo:
Thank you!

Issue #36, opened 2016-10-25
"Denite buffer" order


pocari:
`Unite buffer` enumerates previous buffer as first candidate, and current buffer as last candidate.
I hope so in `Denite buffer`

so, I workaround by these changes.
https://github.com/Shougo/denite.nvim/compare/master...pocari:fix-buffer-source-changes-alternately?expand=1

Is there any way to fix this problem?

Shougo:
LGTM.

Please create new PR for it.
I will merge it.

pocari:
thanks, I created PR!

PR #37, opened 2016-10-25
Fix "Denite buffer" candidates order

pocari:
Unite buffer enumerates previous buffer as first candidate, and current buffer as last candidate.
I hope so in Denite buffer!

related issue #36


Shougo:
Merged.

PR #49, opened 2016-11-04
Port 'isprint' function from Python itself to reduce Vim call.

lambdalisue:
- [x] Lint
- [x] License note


Shougo:
The license agreement is too long.
Please move the function to another file.
Better is in neovim-prompt.

lambdalisue:
It should be in denite's code while neovim-prompt did not port the target code but denite. I'll move it to an isolated file later :-]

2016年11月5日 12:34 +0900 に Shougo <notifications@github.com> は書きました:

lambdalisue:
Done. @Shougo.

Shougo:
LGTM

PR #68, opened 2016-11-19
Fix scroll_up regression

rafi:
Fix `scroll_up` regression caused by 0302bf7f69c19cd0b0cd6183d9e3733ee482405a (_jump to next source_)


Shougo:
Thanks.

PR #69, opened 2016-11-19
Stop buffering on empty receive, not OK

rafi:
- Decouple OK acknowledgement of MPD to client
- Sleep after sending commands to wait
- Fix incorrect exception name


Shougo:
Merged.

PR #71, opened 2016-11-21
Implement sorter_sublime

wh11e7rue:
I have implemented a sorter like Sublime Text's fuzzy matching based on [this article](https://blog.forrestthewoods.com/reverse-engineering-sublime-text-s-fuzzy-match-4cffeed33fdb) and [its implementation](https://github.com/forrestthewoods/lib_fts/blob/master/code/fts_fuzzy_match.js#L33).

I'm using it with init.vim like this:
```vim
  call denite#custom#source('buffer,file_rec,file_mru', 'sorters', ['sorter_sublime'])
```

sorter_sublime:

sorter_rank:

### DIFF
--- a/rplugin/python3/denite/filter/sorter_sublime.py
+++ b/rplugin/python3/denite/filter/sorter_sublime.py
@@ -1,0 +1,124 @@
+from .base import Base
+
+
+class Filter(Base):
+    def __init__(self, vim):
+        super().__init__(vim)
+
+        self.name = 'sorter_sublime'
+        self.description = 'sorter for fuzzy matching like sublime text based on lib_fts'
+
+    def filter(self, context):
+        if len(context['input']) == 0:
+            return context['candidates']
+
+        for candidate in context['candidates']:
+            candidate['filter__rank'] = get_score(context['input'], candidate['word'])
+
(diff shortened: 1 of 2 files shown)


rplugin/python3/denite/filter/sorter_sublime.py
```diff
@@ -0,0 +1,155 @@
+from .base import Base
```
Shougo: Please add the LICENSE.

.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*.py]
+indent_size = 4
```
Shougo: It should be added?

.gitignore
```diff
@@ -1,3 +1,33 @@
...
+.VolumeIcon.icns
+
+# Directories potentially created on remote AFP share
+.AppleDB
+.AppleDesktop
+Network Trash Folder
+Temporary Items
+.apdisk
+
+
```
Shougo: What is this?
Why this is added?

Shougo:
This is the good feature.
But it includes unrelated changes.
Please explain to me.

wh11e7rue:
Sorry...
I have deleted that commit

Shougo:
LGTM

PR #73, opened 2016-11-23
add yank action

pocari:
I have a thing that I want to yank candidate line, so I created this PR.


Shougo:
Thanks.  I will add the document.

PR #76, opened 2016-11-24
Fix yank action register

pocari:
In PR #73, I yanked only `"`" register.  I added `v:register` if has('clipboard'), like `Unite`

### DIFF
--- a/rplugin/python3/denite/kind/base.py
+++ b/rplugin/python3/denite/kind/base.py
@@ -21,4 +21,4 @@ class Base(object):
         self.__yank(self.vim, target['word'])
 
     def __yank(self, vim, word):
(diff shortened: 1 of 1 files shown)


rplugin/python3/denite/kind/base.py
```diff
@@ -22,3 +22,5 @@ def action_yank(self, context):
 
     def __yank(self, vim, word):
         vim.call('setreg', '"', word, 'v')
+        if vim.eval('has("clipboard")') == 1:
```
Shougo: It should be:

```python
if vim.call('has', 'clipboard'):
```

pocari:
fixed

Shougo:
Merged

PR #82, opened 2016-11-26
add wraped cursor option

pocari:
I have implemented `-cursor-wrap` option.

### DIFF
--- a/rplugin/python3/denite/ui/default.py
+++ b/rplugin/python3/denite/ui/default.py
@@ -49,6 +49,7 @@ class Default(object):
                 # Skip the initialization
                 self.__current_mode = context['mode']
                 self.__context['immediately'] = context['immediately']
+                self.__context['cursor_wrap'] = context['cursor_wrap']
 
                 self.init_buffer()
                 self.change_mode(self.__current_mode)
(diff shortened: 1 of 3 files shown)


rplugin/python3/denite/ui/default.py
```diff
@@ -467,7 +468,10 @@ def move_to_next_line(self):
         elif self.__win_cursor + self.__cursor < self.__candidates_len:
             self.__cursor += 1
         else:
-            return
+            if self.__context['cursor_wrap']:
```
Shougo: It should be:

```vim
elif self.__context['cursor_wrap']:
     self.move_to_first_line()
else:
     return

File: rplugin/python3/denite/ui/default.py [hunk 2]
```diff
@@ -476,7 +480,10 @@ def move_to_prev_line(self):
         elif self.__cursor >= 1:
             self.__cursor -= 1
         else:
-            return
+            if self.__context['cursor_wrap']:
```
Shougo: Same above.

pocari:
fixed.

Shougo:
LGTM

PR #83, opened 2016-11-26
fix '-no-' bool option completion

pocari:
If option_name included '_', it is replaced '-' in completion, but In completion with '-no-', it is not replaced.


Shougo:
Thank you

PR #87, opened 2016-11-29
fix foward match to exact match in file open action.

pocari:
Candidates of `file_rec` is followin state

```
aaa
aaa.bbb
```
With `aaa.bbb` was already opened, I chose `aaa`. but `aaa` was not opened and buffer contents kept  `aaa.bbb`.

I refered to https://github.com/vim-jp/issues/issues/990#issuecomment-263584439


Shougo:
Merged

PR #91, opened 2016-11-30
Fix #90

### LINKED ISSUE #90: Action name completion is not working
**Warning:  I will close the issue without the minimal init.vim and the reproduce ways.**

# Problems summary
Action name completion in normal mode is not working, if I canceled anything once confirmed.

## Neovim Python Diagnostic

- Neovim Version: NVIM v0.2.0-1-g9956bee

## Provide a minimal init.vim with less than 50 lines (Required!)

```
let s:dein_dir = expand('~/.cache/dein2')
let s:dein_repo_dir = s:dein_dir . '/repos/github.com/Shougo/dein.vim'

if &runtimepath !~# '/dein.vim'
  if !isdirectory(s:dein_repo_dir)
    execute '!git clone https://github.com/Shougo/dein.vim' s:dein_repo_dir
  endif
  execute 'set runtimepath^=' . fnamemodify(s:dein_repo_dir, ':p')
endif

call dein#begin(s:dein_dir)

call dein#add('Shougo/denite.nvim')

call dein#end()
call dein#save_state()

if dein#check_install()
  call dein#install()
endif
```

## The reproduce ways from neovim starting (Required!)

 1. :Denite file_rec
 2. move `normal mode` by `<C-o>` and `<TAB>`
     I input 'de' and `<TAB>` (expecting 'default').
  
    
   
    and completed 'default' (it's works)

 3. Next, I erase 'default' (by backspace key) and input 'y'
     and input `<TAB>` for completion (expect 'yank')
     but, completion is not working.

lambdalisue:
Copy is required before filter. Fix #90

### DIFF
--- a/autoload/denite/helper.vim
+++ b/autoload/denite/helper.vim
@@ -30,7 +30,7 @@ function! denite#helper#complete(arglead, cmdline, cursorpos) abort "{{{
   return uniq(sort(filter(_, 'stridx(v:val, a:arglead) == 0')))
 endfunction"}}}
 function! denite#helper#complete_actions(arglead, cmdline, cursorpos) abort "{{{
-  return uniq(sort(filter(g:denite#_actions,
+  return uniq(sort(filter(copy(g:denite#_actions),
         \ 'stridx(v:val, a:arglead) == 0')))
 endfunction"}}}
 


Shougo:
Thank you!

PR #95, opened 2016-12-02
Display "No Name" when there is empty buffer name

momo-lab:
Displaying the empty buffer intention in buffer source displays the current directory.
It is displayed as "No Name" like unite.vim.


Shougo:
Thanks

PR #96, opened 2016-12-02
Add syntax highlight for buffer source

momo-lab:
I added highlight syntax of buffer source.
This syntax is almost the same as that of unite.vim.


Shougo:
Thanks

PR #101, opened 2016-12-04
Except 'base' from source name completion

pocari:
I think that it is better to exclude "base" from source name completion.


Shougo:
Thanks

PR #103, opened 2016-12-05
Fix async updating issue.

lambdalisue:
The candidates on the buffer was not updated often enough compared to the previous denite.nvim
Fix that issue by calling 'redraw' internally when 'async' is enabled.


Shougo:
Thank you!

PR #104, opened 2016-12-05
Fix neovim-prompt builtin mapping actions

lambdalisue:
Followed the latest neovim-prompt which fixed the followings

- `<prompt:move_caret_to_one_word_left>` and `<prompt:move_caret_to_one_word_right>`

  It seems the native Vim does not respect `iskeyword` for moving caret by `<S-Left>` or `<S-Right>` so the changes followed the Vim's flavor.
  - [Spec1](https://github.com/lambdalisue/neovim-prompt/blob/0b0d010d558b77b70870d9c924c7834500b16b27/test/test_action.py#L434-L472)
  - [Spec2](https://github.com/lambdalisue/neovim-prompt/blob/0b0d010d558b77b70870d9c924c7834500b16b27/test/test_action.py#L511-L549)

- `<prompt:delete_word_before_caret>`

  Previously the mapping action cannot remove characters which is not listed in `iskeyword`.
  The changes fixed that issue and close the behavior to the one of the native Vim.
  - [Spec](https://github.com/lambdalisue/neovim-prompt/blob/0b0d010d558b77b70870d9c924c7834500b16b27/test/test_action.py#L125-L189)

- `<prompt:delete_word_after_caret>`

  Previously the mapping action cannot remove characters which is not listed in `iskeyword`.
  The changes fixed that issue and close the behavior to the `w` in Normal mode of the native Vim.
  - [Spec](https://github.com/lambdalisue/neovim-prompt/blob/0b0d010d558b77b70870d9c924c7834500b16b27/test/test_action.py#L214-L297)

- `<prompt:delete_word_under_caret>`

  Previously the mapping action cannot remove characters which is not listed in `iskeyword`.
  The changes fixed that issue and close the behavior to the `diw` in Normal mode of the native Vim.
  - [Spec](https://github.com/lambdalisue/neovim-prompt/blob/0b0d010d558b77b70870d9c924c7834500b16b27/test/test_action.py#L308-L391)

### DIFF
--- a/rplugin/python3/denite/prompt/.travis.yml
+++ b/rplugin/python3/denite/prompt/.travis.yml
@@ -4,12 +4,31 @@ python:
   - 3.3
   - 3.4
   - 3.5
+os:
+  - linux
+  - osx
+osx_image: xcode8
+
+addons:
+  apt:
+    packages:
+      - libtool
+      # - libtool-bin # is only required for Ubuntu 16.04/Debian Jessie and newer.
+      - autoconf
+      - automake
+      - cmake
+      - g++
+      - pkg-config
+      - unzip
 
 install:
+  - curl -sL git.io/v18DU | bash
+  - pip install neovim
   - pip install coveralls
   - git clone --depth 1 --single-branch -b ci-test https://github.com/lambdalisue/neovim-prompt ci-test
 
 script:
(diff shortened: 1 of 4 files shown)


Shougo:
Thank you.

PR #107, opened 2016-12-08
Fix test for existence of keyword argument in `parse_command`

ohcibi:
The test was like `kwargs[arg]` which throws an exception if there is no
keyword argument named `arg` instead of returning false. `arg in kwargs`
will return false and thus work in an if-clause.

This commit also uses a more idiomatic approach on returning the parsed
command in the end.

I noticed that after the other PR was merged, so I needed to create a
new one.


Shougo:
Thank you.

PR #110, opened 2016-12-11
Fix action branch for a single source.

lambdalisue:
Without this PR on `action` branch does not work when a single source has specified.

## Procedure

1. `Denite file_rec`
2. `<C-o>`
3. `<Space>`

will do nothing.

With this PR, single source like above works like

selection_056

### DIFF
--- a/rplugin/python3/denite/ui/default.py
+++ b/rplugin/python3/denite/ui/default.py
@@ -23,6 +23,14 @@ class Default(object):
     def current_mode(self):
         return self.__current_mode
 
+    @property
+    def __selected_icon(self):
+        return self.__context['selected_icon']
+
+    @property
+    def __short_source_names(self):
+        return self.__context['short_source_names']
+
     def __init__(self, vim):
         self.__vim = vim
         self.__denite = denite.Denite(vim)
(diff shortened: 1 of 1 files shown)


rplugin/python3/denite/ui/default.py
```diff
@@ -246,23 +254,32 @@ def update_buffer(self):
...
 
         self.__options['modified'] = False
 
         self.move_cursor()
 
+    def __get_candidate_display_text(self, index):
+        candidate = self.__candidates[index]
+        terms = []
+        if index in self.__selected_candidates:
+            terms.append(self.__selected_icon)
```
lambdalisue: I don't like to access a member of dictionary directly so I add a new private property. Let me know if you don't like this way.

rplugin/python3/denite/ui/default.py
```diff
@@ -246,23 +254,32 @@ def update_buffer(self):
...
+        if self.__is_multi:
+            if self.__short_source_names:
```
lambdalisue: Same.

Shougo:
Thank you!

PR #111, opened 2016-12-13
Fix menu action and add description

rafi:
- Fix action used to open menu
- Add description to candidate word

PR #115, opened 2016-12-14
Fix menu action and add description

rafi:
⚠️ Seems that #111 wasn't properly merged.

- Fix action used to open menu
- Add description to candidate word


Shougo:
OH, OK

PR #119, opened 2016-12-15
Fix not recognizing different options

tjdevries:
When I had, for example, a menu, where some menus were commands and some menus were files, it did not work.

The problem was that the iteration was not actually selecting the right items. I fixed the typo.


tjdevries:
This is the reason that the fixes for #115 were added. They weren't actually necessary, it was this bug that was preventing the menu source from working correctly.

Shougo:
Thank you.

PR #123, opened 2016-12-17
source: Fix and document unintuitive behavior, Fixes #113

### LINKED ISSUE #113: source `grep`: context `input` and `path` non-functional
- [Always have to manually press `<CR>` to confirm `input('Pattern: ')`](https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/source/grep.py#L57-L58)
- Narrowing is always applied resulting in extremely bad performance when filter is applied
- [Can't specify `path` and `args` at the same time, due to `directory` being overwritten by `args[0]`](https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/source/grep.py#L50-L54)

Works great with Unite:
```
function! custom#VisualGrep(options) abort
    call unite#start(['grep'], {
        \ 'path': getcwd(),
        \ 'input': custom#VisualGet(),
        \ 'custom_grep_default_opts': printf('%s %s', g:unite_source_grep_default_opts, a:options),
    \ })
endfunction
```

Denite:
```
function! custom#VisualGrep(...) abort
    let l:options = get(a:000, 0, [])
    let l:args = [getcwd()] + l:options
    call denite#start([{'name': 'grep', 'args': l:args}], {'input': custom#VisualGet(), 'mode': 'normal'})
endfunction
```

nfnty:
Fix and document unintuitive behavior for `args` and `path`.

`grep` pattern applies without inserting narrowing text. Pattern always interpreted literally; fixes problems parsing `-`.

Fixes #113

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -142,19 +142,21 @@ EXAMPLES						*denite-examples*
 
 	" Ack command on grep source
 	call denite#custom#var('grep', 'command', ['ack'])
-	call denite#custom#var('grep', 'recursive_opts', [])
-	call denite#custom#var('grep', 'final_opts', [])
-	call denite#custom#var('grep', 'separator', [])
 	call denite#custom#var('grep', 'default_opts',
 			\ ['--ackrc', $HOME.'/.ackrc', '-H',
 			\ '--nopager', '--nocolor', '--nogroup', '--column'])
+	call denite#custom#var('grep', 'recursive_opts', [])
+	call denite#custom#var('grep', 'pattern_opt', ['--match'])
+	call denite#custom#var('grep', 'separator', ['--'])
+	call denite#custom#var('grep', 'final_opts', [])
 	" Ripgrep command on grep source
 	call denite#custom#var('grep', 'command', ['rg'])
-	call denite#custom#var('grep', 'recursive_opts', [])
-	call denite#custom#var('grep', 'final_opts', [])
-	call denite#custom#var('grep', 'separator', ['--'])
 	call denite#custom#var('grep', 'default_opts',
 			\ ['--vimgrep', '--no-heading'])
+	call denite#custom#var('grep', 'recursive_opts', [])
+	call denite#custom#var('grep', 'pattern_opt', ['--regexp'])
+	call denite#custom#var('grep', 'separator', ['--'])
+	call denite#custom#var('grep', 'final_opts', [])
 
 	call denite#custom#source('file_mru', 'converters',
 	      \ ['converter_relative_word'])
@@ -884,8 +886,8 @@ SOURCES							*denite-sources*
 						*denite-source-buffer*
 buffer		Gather buffers and jump to the buffer.
 
-		Source arguments:
-		1. '!' : a flag to show unlisted buffers
+		Source args:
+			0. '!' : a flag to show unlisted buffers
 
 		Source custom variables:
 		date_format	  the format string to show date
@@ -904,8 +906,8 @@ directory_rec	Gather directories recursive and nominates all directory names
 		under the search directory (argument 1) or the current directory
 		(if argument is omitted) as candidates.
 
-		Source arguments:
-		1. the search directory.
+		Source args:
+			0. the search directory
 
 		Source custom variables:
 		command		 the default get directories command
@@ -925,8 +927,8 @@ file_rec	Gather files recursive and nominates all file names under the
 		search directory (argument 1) or the current directory (if
 		argument is omitted) as candidates.
 
-		Source arguments:
-		1. the search directory.
+		Source args:
+			0. the search directory
 
 		Source custom variables:
 		command		the default get files command.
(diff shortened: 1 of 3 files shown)


Shougo:
Oh, too long diff...
It takes much time to review it.

nfnty:
Just review each commit independently, they're logically laid out.

nfnty:
Needs to be modified slightly due to ddbb9a8fd16f0e5a6a2241d6824928dc2c97707e

doc/denite.txt
```diff
@@ -142,19 +142,21 @@ EXAMPLES						*denite-examples*
...
-	call denite#custom#var('grep', 'final_opts', [])
-	call denite#custom#var('grep', 'separator', [])
 	call denite#custom#var('grep', 'default_opts',
 			\ ['--ackrc', $HOME.'/.ackrc', '-H',
 			\ '--nopager', '--nocolor', '--nogroup', '--column'])
-	" Ripgrep command on grep source
-	call denite#custom#var('grep', 'command', ['rg'])
 	call denite#custom#var('grep', 'recursive_opts', [])
-	call denite#custom#var('grep', 'final_opts', [])
+	call denite#custom#var('grep', 'pattern_opt', ['--match'])
```
Shougo: Why the `pattern_opt` is needed?
Please explain to me.

doc/denite.txt
```diff
@@ -949,15 +951,15 @@ command		Gather commands and put on ex-mode.
...
 		Note: The command result must be "path:linenr:text" pattern.
-		Note: If the input contains the spaces, you must quote it.
 
-		Source arguments:
-		1. the search directory.
-		2. the additional arguments.
+		Source args (all optional):
+			0. list of additional arguments
+			1. list of patterns
+			2. list of paths
```
Shougo: Source arguments must not be List.
It must be String.  Please fix them.

rplugin/python3/denite/source/grep.py
```diff
@@ -40,23 +52,26 @@ def __init__(self, vim):
...
-            directory = context['args'][0]
-        if not directory:
-            directory = context['path']
-        context['__arguments'] = context['args'][1:]
-        context['__directory'] = self.vim.call('expand', directory)
-        context['__input'] = context['input']
-        if not context['__input']:
-            context['__input'] = self.vim.call('input', 'Pattern: ')
+        try:
+            context['__arguments'] = context['args'][0]
```
Shougo: It is very ugly.  I will convert `context['args']` to Dictinary.
Please use `get()`.

Shougo:
Why grep source behavior is changed?
Please read #17 and #19.

@carlitux Please review the changes.  It changes grep behavior.

rplugin/python3/denite/source/file_rec.py
```diff
@@ -25,18 +25,19 @@ def __init__(self, vim):
...
     def on_init(self, context):
-        self.__proc = None
-        directory = context['args'][0] if len(
-            context['args']) > 0 else context['path']
-        context['__directory'] = self.vim.call('expand', directory)
+        context['__proc'] = None
+        try:
+            context['__directory'] = context['args'][0]
+        except IndexError:
+            context['__directory'] = self.vim.call('getcwd')
```
Shougo: Please don't change the current behavior.
Where is `self.vim.call('expand', directory)`?
It is needed.

Shougo: And please don't use `self.vim.call('getcwd')`.

Shougo:
I have reviewed.  Many problems are exists.
I cannot merge.

doc/denite.txt
nfnty: Ability to specify multiple patterns. Ability to specify patterns starting with `-`.

doc/denite.txt
nfnty: Why? Would a dict of lists be better? Must be able to specify multiple arguments.

Shougo: Because, it cannot be set by the command line.
And you don't check the types in the souce.
Have you tested?

rplugin/python3/denite/source/grep.py
Shougo: Please use below:

```python
args = dict(enumerate(context['args']))
context['__arguments'] = args.get(0, ...)
```

doc/denite.txt
nfnty: Then can there be a decoupling between `Denite` and `denite#start()`? I need the ability to specify arguments in a more complex way than a string. That includes using unmodified paths, not expanded with `expand()`.

nfnty:
I could possibly implement a compatibility layer.

#19 makes no sense... Why would you make inputting patterns extremely tedious when that's what input is intended for? If you want other options then create a new alias.

nfnty:
I'll split this PR.

rplugin/python3/denite/source/file_rec.py
Shougo: `directory_rec.py` is bug.  I will fix it.
`file_point.py` is the feature.

doc/denite.txt
Shougo: Your source does support List only.
It is broken.
You can support List and String types.

Shougo:
Please fix the conflicts.

nfnty:
Fixed. Full backwards compatibility except:

Input is no longer split, instead user has to append the arguments on the command line like:
```
:Denite grep::--ignore-case\ --fixed-strings
```

Shougo:
I have checked it locally.  Good.

PR #124, opened 2016-12-19
Fix buffer syntax highlight

momo-lab:
Fix to enable highlight of deniteSource_buffer_Time.

before
before

after
after


Shougo:
OK

PR #127, opened 2016-12-19
Fix error when nesting Denite command

momo-lab:
Fix a problem that nested Denite command caused the following error.
It occurred in vim8 only.

It occurs when you input ``:Denite menu<CR><CR><ESC>``
```
Error detected while processing function denite#helper#call_denite[14]..denite#start[23]..denite#vim#_start:
line   23:
Traceback (most recent call last):
  File "<string>", line 21, in <module>
NameError: name '_temporary_scope' is not defined
```


Shougo:
LGTM

PR #128, opened 2016-12-19
Add 'ex' action to execute a 'word' as a command

lambdalisue:
This PR implement an 'ex' action which was in unite.vim.
Personally, I have not used 'ex' action so the behavior may be not correct. In that case, let me know where I should fix.

@momo-lab Could you try and give me a feedback?

### DIFF
--- a/rplugin/python3/denite/kind/base.py
+++ b/rplugin/python3/denite/kind/base.py
@@ -35,6 +35,9 @@ def _yank(vim, word):
 
 
 def _ex(vim, word):
-    expr = vim.call('input', ':', word, 'command')
+    # NOTE:
+    # <C-b> (\x02) in a command-line move the caret to the beginning.
+    # Somehow the key above works in 'input()' function as well.
+    expr = vim.call('input', ':', ' %s\x02' % word, 'command')
     if expr:
         vim.command(expr)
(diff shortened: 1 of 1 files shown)


momo-lab:
I tried it.
I want the initial cursor position to start.
How about this fix?(^B inputted by \<C-v\>\<C-b\>)

```diff
diff --git a/rplugin/python3/denite/kind/base.py b/rplugin/python3/denite/kind/base.py
index 6a8cd6e..067ff3d 100644
--- a/rplugin/python3/denite/kind/base.py
+++ b/rplugin/python3/denite/kind/base.py
@@ -35,6 +35,6 @@ def _yank(vim, word):
 
 
 def _ex(vim, word):
-    expr = vim.call('input', ':', word, 'command')
+    expr = vim.call('input', ':', ' {}^B'.format(word), 'command')
     if expr:
         vim.command(expr)
```

lambdalisue:
Thaks. I'll try 👍  But I don't know which should be a correct behavior. Could you tell me the way I should follow @Shougo?

lambdalisue:
@momo-lab Fixed. I've changed `^B` to `\x02` while `^B` is a special character.

momo-lab:
Thanks! 👍

Shougo:
Looks good.

PR #132, opened 2016-12-20
[Feature Request][WIP] Add `jump` action to jump opened buffer window

pocke:
Hi

I need a new action for jumping to opened window for buffer source.

Action behavior
------

1. select a buffer by buffer source
1. get window number by `win_findbuf()` function.
    - If window exists, jump to the window by `win_gotoid()`
    - If window doesn't exist, fallback other action (Note: I've implement fallback to tabopen action, but I think the action should be selectable by user. But I don't know how to implement it.)

dac

Why work in progress
------

As mentioned above, user can't select the fallback action in this implementation. 
Could you tell me how to implement it?

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -1106,6 +1106,22 @@ openable	An interface to open the object.
 			vsplit
 				open the object with 'vsplit'.
 
+			switch
+				Switch to the window that is opening the
+				object.
+
+			tabswitch
+				Switch to the window that is opening the
+				object. If it isn't opened, fallback tabopen.
+
+			splitswitch
+				Switch to the window that is opening the
+				object. If it isn't opened, fallback split.
+
+			vsplitswitch
+				Switch to the window that is opening the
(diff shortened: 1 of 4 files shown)


Shougo:
> I need a new action for jumping to opened window for buffer source.

It should be `switch` action instead of `jump`.

Shougo:
> As mentioned above, user can't select the fallback action in this implementation.
Could you tell me how to implement it?

I think the default `tabopen` behavior is bad.
It should be `open` action.
But denite has not action customization support.

Shougo:
unite.vim has `switch`, `tabswitch`, `splitswitch`, `vsplitswitch` actions.

pocke:
Thanks for your help! I've implemented the `switch`, `tabswitch`, `splitswitch` and `vsplitswitch` features.
It works with buffer and file kind.

pocke:
Sorry, I forgot to write the document. I'll write it now

Shougo:
LGTM

PR #135, opened 2016-12-21
Create a new file if it's not in the search results

nicholasday:
`file_rec_new.py` is a copy of `file_rec.py`, but it changes the sorter to `sorter_rank_new.py`. `sorter_rank_new.py` adds a candidate if the input is not '' at the end of the candidates list. Please let me know if there's anything I need to fix.


Shougo:
Yes, this feature is needed.
But denite.nvim should support to create new candidates feature.
So I cannot accept your changes.  Sorry.

nicholasday:
I really appreciate your work. Thanks for what you bring to the (n)vim community!

Issue #56, opened 2016-11-09
"Denite grep", quote required?


pocari:
**Warning:  I will close the issue without the minimal init.vim and the reproduce ways.**

# Problems summary
In `Denite grep`, Using search word separated space, error has occured in `process.py`  

`test.txt` contents is
```
Dark powered for Nvim
Dark powered for Vim8
```

and I run `Denite grep` , entered `dark powerd`.

I logged the arguments passed to `subprocess.Popen` in `Process#__init__`.
and I run it on IPython repl. 

So, following error.
```
In [9]: command2
Out[9]: ['grep', '-inH', '-r', '--', 'dark', 'powered', '.']

In [10]: subprocess.Popen(command2)
Out[10]: grep: powered: No such file or directory
./test.txt:1:Dark powered for Nvim
./test.txt:2:Dark powered for Vim8
<subprocess.Popen at 0x10f811048>
```
 use only `dark`,  ignored `powered`.

If I enterd `'dard powered'`, Arguments passed to `subprocess.Popen` is `['grep', '-inH', '-r', '--', 'dark powered', '.']`

```
In [4]: command
Out[4]: ['grep', '-inH', '-r', '--', 'dark powered', '.']

In [5]: subprocess.Popen(command)
./test.txt:1:Dark powered for Nvim
./test.txt:2:Dark powered for Vim8
Out[5]: <subprocess.Popen at 0x10e639d30>
```

It seems good.

## Expected

Input patterns should be quoted automatically(?)

## Provide a minimal init.vim with less than 50 lines (Required!)

`test_vimrc`
```vim
let s:dein_dir = expand('~/.cache/dein2')
let s:dein_repo_dir = s:dein_dir . '/repos/github.com/Shougo/dein.vim'

if &runtimepath !~# '/dein.vim'
  if !isdirectory(s:dein_repo_dir)
    execute '!git clone https://github.com/Shougo/dein.vim' s:dein_repo_dir
  endif
  execute 'set runtimepath^=' . fnamemodify(s:dein_repo_dir, ':p')
endif

call dein#begin(s:dein_dir)

call dein#add('Shougo/denite.nvim')

call dein#end()
call dein#save_state()

if dein#check_install()
  call dein#install()
endif
```

## The reproduce ways from neovim starting (Required!)

```
% cat test.txt
Dark powered for Nvim
Dark powered for Vim8
```

nvim -u test_vimrc
 
:Denite grep
dark powered

Shougo:
[...]

PR #137, opened 2016-12-25
Improve grep to handle extra arguments on pattern

carlitux:
This allow to create better queries on ack or ag.

### DIFF
--- a/rplugin/python3/denite/source/grep.py
+++ b/rplugin/python3/denite/source/grep.py
@@ -142,7 +142,7 @@ class Source(Base):
         commands += self.vars['recursive_opts']
         commands += context['__arguments']
         for pattern in context['__patterns']:
-            commands += self.vars['pattern_opt'] + [pattern]
+            commands += self.vars['pattern_opt'] + shlex.split(pattern)
         commands += self.vars['separator']
         if context['__paths']:
             commands += context['__paths']


Shougo:
@nfnty Please check it.

Shougo:
@carlitux Please describe why this feature is needed to @nfnty.

carlitux:
@Shougo @nfnty well with ack or ag I have a mapping to call 

```vim
:Denite grep
```

but on large projects I want to filter searches adding --js or --css or any other type defined but inside the pattern input.

nfnty:
This is stupid. If you need extra arguments just call: `:Denite grep::--js`

And if it's too much effort to type that, then just map: `nnoremap <Leader>g :<C-u>Denite grep::` and just type in your arguments and press enter.

carlitux:
why this could be stupid?

- @Shougo's plugins tries to be flexible so why not give that flexibility to input pattern?
- Why should I have more than one mapping to get quickly the input pattern? and type once enter key instead of typing twice to get the result?
- I am not sure what does "-e" in grep or "--match" in ack but why not allow developer to specify it or any other flags that are often needed. https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/source/grep.py#L73 force create compatibility code.
- On input pattern now also we can't specify quickly directory where search.
- :Denite grep::[args] should be used to our base arguments then allow us add others to those in the input pattern

nfnty:
Because you're fucking with everyone's patterns just because of your special use case. Regex patterns can be very complex and you don't want to make it harder than it already is.

Yes @Shougo makes very flexible plugins, due to the fact that he tries to eliminate stupid design decisions that will hold everyone back.

If my above mapping is too much effort then you can easily accomplish exactly what you want with a simple function:
```
function! Grep() abort
    python3 << EOF
import vim, shlex
args = shlex.split(vim.call('input', 'Arguments: '))
if len(args) >= 1:
    vim.call('denite#start', [{'name': 'grep', 'args': [[], args[:-1], [args[-1]]]}],
        {'path': vim.call('getcwd'), 'mode': 'normal'})
EOF
endfunction
```

Seriously read `:help Denite` if you don't know how to write custom functions and don't want to use the builtin commands.

lambdalisue:
**I don't think this is a stupid idea**. I understand that sometime users wonder if they can specify some options in a `Pattern:` prompt so the suggestions above are bit out of the point from the request I guess.

However I don't feel the PR should be merged neither while this approach has the following issues:

1. Currently, `Denite grep` highlights matched text by Python's regex built from `context['patterns']` so if options like `--js` were allowed in the pattern, it makes things difficult to distinguish the patterns or the options

2. Even if `Denite grep` does not use `context['patterns']` for highlighting, allowing options in patterns increase complexity and reduce flexibility. It would make implementing new features difficult 

Alternative idea for the request is using a customizable prompt rather than a builtin `input()` for the `Pattern:` prompt and add a custom mapping (say, `Ctrl-^`) to change the prompt target from the pattern to the options. Then user can modify options rather than pattern by hitting `Ctrl-^`. 
However, **the implementation would be complex and I don't think it worth to do that**, sorry.

Well, I personally don't like to see someone being called stupid or fxxxin'.

nfnty:
No one was called stupid. I called this implementation stupid for fucking with everyone's patterns for a very specific use case and as is would make the plugin unusable. This can easily be fixed with custom functions/mappings.

I even went out of my way to help him out; just use the function I came up with in my previous comment and start fixing bigger problems.

Shougo:
Closed.
It has the problems and #56.

Shougo:
@nfnty 

> I am not sure what does "-e" in grep or "--match" in ack but why not allow developer to specify it or any other flags that are often needed. https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/source/grep.py#L73 force create compatibility code.

By the way, what do you mean the code?
I don't want to add the ugly hack.

PR #142, opened 2016-12-31
preserve exit mode for resume

chemzqm:
When `:Denite -resume` is used the mode would always change to `insert`,  I think it's better to preserve the mode.


Shougo:
But I think it cannot change the current mode by the arguments.

Shougo:
Fixed.

PR #145, opened 2017-01-05
Do not override builtin functions of Python 3

lambdalisue:
Note: The followings were builtin functions of Python2 but Python3 so ignored

- 'file'
- 'buffer' (This was a non-essential builtin functions even in Python2)


Shougo:
Merged.

PR #146, opened 2017-01-06
Check if kind a is Kind object in get_actions

chemzqm:
`get_actions` would throw error if the kind in source is a kind object.


Shougo:
Thank you.

PR #147, opened 2017-01-06
Add redraw for delete action

chemzqm:
I have a file most recently used plugin, I've add the delete_action to that kind, the problem is I don't have UI object reference in the Kind object, so I can't make it redraw, these code solves that by redraw after delete action.

### DIFF
--- a/rplugin/python3/denite/denite.py
+++ b/rplugin/python3/denite/denite.py
@@ -235,6 +235,7 @@ class Denite(object):
             return {}
         return {
             'name': action_name,
+            'is_redraw': action_name == 'delete',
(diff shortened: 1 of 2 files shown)


Shougo:
Thank you.

> I have a file most recently used plugin

Why does not use `file_old` source?

chemzqm:
@Shougo https://github.com/chemzqm/redismru.vim it could support thousands of files and support sync files of different vim session

Shougo:
> it could support thousands of files and support sync files of different vim session

You can change the value by `'viminfo'` option.
And the sessions are merged in Vim 8.0/neovim.

PR #160, opened 2017-01-20
Add winminheight for auto-resize windows

mackeybj:
Small patch to add minimum height for autoresizing windows. This stops the window from having to redraw its size beyond a certain height when the candidate list is shrinking as I personally find this somewhat distracting.

The naming is the same as the vim/neovim's minimum window height for consistency.

> The 'winheight' option can be set to a minimal desired height of a window and
> 'winminheight' to a hard minimum height.
>    Likewise, there is 'winwidth' for the minimal desired width and
> 'winminwidth' for the hard minimum width.
>    The 'equalalways' option, when set, makes Vim equalize the windows sizes
> when a window is closed or opened.


Shougo:
LGTM

PR #163, opened 2017-01-21
Stop escaping colon in s:eval_cmdline

TKNGUE:
s:eval_cmdlin escapes colon after adapting eval() to context.
But this might be leaded into error.

Test Case:
   context  Denite `has('nvim') ? file_rec:. : file_rec`


Shougo:
OK.

PR #174, opened 2017-02-06
Use relative for open when possible

chemzqm:
When I open a file with `Denite file_rec` and the file is not associated with existing buffer, a full path would be shown on my statusline, like: ` ~/vim-dev/unite-extra/autoload/unite/sources/node.vim`
That would confuse me since I would think of the file is not inside cwd.

### DIFF
--- a/rplugin/python3/denite/kind/file.py
+++ b/rplugin/python3/denite/kind/file.py
@@ -5,6 +5,7 @@
 # ============================================================================
 
 import re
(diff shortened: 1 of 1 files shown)


Shougo:
Please fix the Travis error.

chemzqm:
@Shougo Fixed

Shougo:
LGTM

PR #187, opened 2017-02-21
Create scantree.py file that can be used as command for file_rec on windows

skeept:
When I try to use file_rec on windows 64 nothing is shown.
I experimented with the provided path and it seems to work.

### DIFF
--- a/rplugin/python3/denite/scantree.py
+++ b/rplugin/python3/denite/scantree.py
@@ -1,29 +1,47 @@
+# ============================================================================
+# FILE: scantree.py
+# AUTHOR: Jorge Rodrigues <skeept at gmail.com>
+# License: MIT license
+# ============================================================================
+
 import sys
-import os
+from os.path import basename, join
+from os import walk
 
+try:
+    from os import scandir2
+    def scantree(path_name, skip_list=None):
+        """if scandir is available use it as it is faster than os.walk"""
 
-def scantree(pathn):
-    """Use scandir if available else os.walk.
-    scandir is supposed to be much faster"""
+        if skip_list is None:
+            skip_list = ['.git', '.hg']
 
-    skip_list = ['.git']
-
-    if 'scandir' in dir(os):
-        for entry in os.scandir(pathn):
+        for entry in scandir(path_name):
             if entry.is_dir(follow_symlinks=False):
-                if os.path.basename(entry.path) not in skip_list:
+                if basename(entry.path) not in skip_list:
                     yield from scantree(entry.path)
             else:
                 yield entry.path
-    else:
-        for root, dirn, files in os.walk(pathn):
-            if os.path.basename(root) in skip_list:
+except ImportError:
+    def scantree(path_name, skip_list=None):
+        """scandir is not available use os.walk"""
+
+        if skip_list is None:
+            skip_list = ['.git', '.hg']
+
+        for root, dirn, files in walk(path_name):
+            if basename(root) in skip_list:
                 continue
             for fn in files:
-                yield os.path.join(root, fn)
+                yield join(root, fn)
+
+
+def output_files():
+    """print the list of files to stdout"""
+    for path_name in sys.argv[1:]:
+        for fn in scantree(path_name):
+            sys.stdout.write(fn + '\n')
 
 
 if __name__ == "__main__":
-    for pathn in sys.argv[1:]:
-        for fn in scantree(pathn):
-            sys.stdout.write(fn + '\n')
+    output_files()
--- a/rplugin/python3/denite/scantree.py
+++ b/rplugin/python3/denite/scantree.py
@@ -10,6 +10,7 @@ from os import walk
 
 try:
     from os import scandir
+    
     def scantree(path_name, skip_list=None):
         """if scandir is available use it as it is faster than os.walk"""
 
--- a/rplugin/python3/denite/scantree.py
+++ b/rplugin/python3/denite/scantree.py
@@ -8,14 +8,23 @@ import sys
 from os.path import basename, join
 from os import walk
 
+
+DEFAULT_SKIP_LIST = ['.git', '.hg']
+
 try:
+    # 'scandir' was introduced at Python 3.5.
+    # See https://pypi.python.org/pypi/scandir
     from os import scandir
 
     def scantree(path_name, skip_list=None):
-        """if scandir is available use it as it is faster than os.walk"""
+        """This function returns the files present in path_name, including the
+        files present in subfolders.
+
+        Implementation uses scandir, if available, as it is faster than
+        os.walk"""
 
         if skip_list is None:
-            skip_list = ['.git', '.hg']
+            skip_list = DEFAULT_SKIP_LIST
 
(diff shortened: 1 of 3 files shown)


rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,29 @@
...
+import os
+
+
+def scantree(pathn):
+    """Use scandir if available else os.walk.
+    scandir is supposed to be much faster"""
+
+    skip_list = ['.git']
+
+    if 'scandir' in dir(os):
```
lambdalisue: Do

```
try:
    from os import scandir
    # scantree with scandir
except ImportError:
    # scantree without scandir
    def scantree(path):
        for root, ....
```

at the very top of the module and re-write entire you code with pre-deined `scandir`. Executing `dir(os)` everytime is not good for the performance.

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,29 @@
+import sys
+import os
+
+
+def scantree(pathn):
```
lambdalisue: `path`?

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,29 @@
...
+            if os.path.basename(root) in skip_list:
+                continue
+            for fn in files:
+                yield os.path.join(root, fn)
+
+
+if __name__ == "__main__":
+    for pathn in sys.argv[1:]:
+        for fn in scantree(pathn):
+            sys.stdout.write(fn + '\n')
```
lambdalisue: Remove this kind of unittest code. Write real test code instead if you need (I'm not sure denite.nvim already have an official way though)

Shougo:
@skeept Please add the documentation in file_rec source in deoplete.txt.

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,29 @@
+import sys
```
Shougo: Please add the LICENSE header file.

rplugin/python3/denite/scantree.py
skeept: I will follow your suggestion but right now the script is called every time with python so that only gets executed once.

Maybe in the future it can be imported and used directly and in that case it makes sense to care a little more about the performance.

rplugin/python3/denite/scantree.py
skeept: This code actually needs to run because we need to print the list of files to output. I create a function and call it to make more explicit.

I still think it makes sense to keep this under the __name__ == "__main__" check.

skeept:
In fact I would prefer not to call an external command to be able to get the files in linux.
I would prefer to just create a thread and use that in a async way.
But I don't know how to add candidates from calling a python function in a thread I just followed the examples present here.

The change for this would be based on the following code:

```python
from threading import Thread
from queue import Queue
import os
import sys
import time


def scantree_thread(folder):
    def scantree(path_name, outq):
        batch = []
        max_size = 10
        for root, dirn, files in os.walk(path_name):
            for fn in files:
                batch.append(os.path.join(root, fn))
                if len(batch) >= max_size:
                    outq.put(tuple(batch))
                    batch = []
        if batch:
            outq.put(tuple(batch))


    out_q = Queue()
    tr = Thread(target=scantree, args=(folder, out_q))
    tr.start()

    while tr.isAlive():
        if out_q.qsize() > 0:
            batch = out_q.get()
            for fn in batch:
                yield fn
        time.sleep(0.001)

    tr.join()
```

I don't know how to go from this code to actually add the candidates to the list...

Shougo:
@lambdalisue Please review the changes.

lambdalisue:
Sorry. I probably mis-understood the current denite impl. I'll review it later

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,48 @@
...
+    def scantree(path_name, skip_list=None):
+        """scandir is not available use os.walk"""
+
+        if skip_list is None:
+            skip_list = ['.git', '.hg']
+
+        for root, dirn, files in walk(path_name):
+            if basename(root) in skip_list:
+                continue
+            for fn in files:
```
lambdalisue: It is a bit personal reason but I feel `fn` is a short form of `function` so `filename` is better in this case.

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,48 @@
...
+
+import sys
+from os.path import basename, join
+from os import walk
+
+try:
+    from os import scandir
+
+    def scantree(path_name, skip_list=None):
+        """if scandir is available use it as it is faster than os.walk"""
```
lambdalisue: Well, this is kind a personal reason but the document comment should explain what the function stands for. This kind of comment should be in after the summary line or a normal command.

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,48 @@
...
+# AUTHOR: Jorge Rodrigues <skeept at gmail.com>
+# License: MIT license
+# ============================================================================
+
+import sys
+from os.path import basename, join
+from os import walk
+
+try:
+    from os import scandir
```
lambdalisue: It would be nice to comment

```
# 'scandir' has introduced at Python 3.5. See https://pypi.python.org/pypi/scandir
```

For users who do not know about `scandir`

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def scantree(path_name, skip_list=None):
+        """if scandir is available use it as it is faster than os.walk"""
+
+        if skip_list is None:
+            skip_list = ['.git', '.hg']
```
lambdalisue: This value is shared with `scantree` without `scandir` so I suggest you to define `DEFAULT_SKIP_LIST` on a module level to make two function compatible.

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for entry in scandir(path_name):
+            if entry.is_dir(follow_symlinks=False):
+                if basename(entry.path) not in skip_list:
+                    yield from scantree(entry.path)
+            else:
+                yield entry.path
+except ImportError:
+    def scantree(path_name, skip_list=None):
+        """scandir is not available use os.walk"""
```
lambdalisue: Same.

rplugin/python3/denite/scantree.py
```diff
@@ -0,0 +1,48 @@
...
+
+        if skip_list is None:
+            skip_list = ['.git', '.hg']
```
lambdalisue: Same.

rplugin/python3/denite/scantree.py
skeept: I tried to satisfy your suggestions. I want to follow your advice as much as possible as you seem to be more experienced in this matter than I am :)

lambdalisue:
Cool thanks 👍

Shougo:
Merged.

PR #188, opened 2017-02-22
Documentation for Kind

chemzqm:
It's useful for custom source.

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -1451,12 +1451,67 @@ abbr		(String)			(Optional)
 ==============================================================================
 CREATE KIND						*denite-create-kind*
 
-Note: Not implemented yet...
+Custom Kind can be created to contain extra actions for your custom source.
+Before create your new kind, you should have a look at existing kinds in
+rplugin/python3/denite/kind/*.py
+
+Instead of create new kind from scratch, you can extend one of the existing
+Kind class, let's take "openable" for example: >
+
+    from ..kind.openable import Kind as Openable
+
+    class Kind(Openable):
+        def __init__(self, vim):
+            super().__init__(vim)
+            self.name = 'mykind'
+
+        def action_open(self, context):
+            target = context['targets'][0]
+            path = target['action__path']
+            # assume you have a MyEdit command
+            self.vim.call('MyEdit', path)
+>
+After create new Kind class, alias it to your source by >
+
+    self.kind = Kind(vim)
+<
+in the "__init__" method of your source class.
 
 ------------------------------------------------------------------------------
 KIND ATTRIBUTES					*denite-kind-attributes*
 
-Note: Not implemented yet...
+					*denite-kind-attribute-__init__*
+__init__	(Function)
+		Kind constructor.  It is always called in initializing.  It
+		must call super() constructor.  This function takes {self} and
+		{vim} as its parameters.
+
+					*denite-kind-attribute-name*
+name 		(String) 			(Required)
+		The name of a Kind.
+
+					*denite-kind-attribute-default_action*
+default_action  (String)
+		Default action name.
+
+					*denite-kind-attribute-persist_actions*
+persist_actions (List) 				(Optional)
+		List of action names which not close Denite buffer after
+		executed.
+
+					*denite-kind-attribute-redraw_actions*
+redraw_actions (List) 				(Optional)
+		List of action names which redraw Denite buffer after
+		executed.
+
+					*denite-kind-attribute-action_*
+action_{name} (Function) 			(Optional)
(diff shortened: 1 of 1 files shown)


doc/denite.txt
```diff
@@ -1451,12 +1451,64 @@ abbr		(String)			(Optional)
 ==============================================================================
 CREATE KIND						*denite-create-kind*
 
-Note: Not implemented yet...
+Custom Kind can be created to contain extra actions for your custom source.
+Before create your new kind, you should have a look at existing kinds in
+rplugin/python3/denite/kind/*.py
+
+Instead of create new kind from sketch, you can extend one of the exiting
```
nelstrom: To my ear, "from sketch" sounds weird. Did you mean "Instead of creating a new kind *from scratch"*?

e-kwsm: s/exiting/existing/

chemzqm: Yep, I've picked wrong word 😄

chemzqm: thanks

doc/denite.txt
```diff
@@ -1451,12 +1451,64 @@ abbr		(String)			(Optional)
...
+Before create your new kind, you should have a look at existing kinds in
+rplugin/python3/denite/kind/*.py
+
+Instead of create new kind from scratch, you can extend one of the existing
+Kind class, let's take `openable` for example: >
+
+	from ..kind.openable import Kind as Openable 
+
+	class Kind(Openable):
+		def __init__(self, vim):
```
lambdalisue: It is an example code for Python. Why do you use 8 spaces (or tab?)

doc/denite.txt
```diff
@@ -1451,12 +1451,64 @@ abbr		(String)			(Optional)
...
+			super().__init__(vim)
+			self.name = 'mykind'
+
+		def action_open(self, context):
+			target = context['targets'][0]
+			path = target['action__path']
+			" assume you have a MyEdit command
```
lambdalisue: Wrong comment char. The line should starts from `#`

doc/denite.txt
```diff
@@ -1451,12 +1451,64 @@ abbr		(String)			(Optional)
 ==============================================================================
 CREATE KIND						*denite-create-kind*
 
-Note: Not implemented yet...
+Custom Kind can be created to contain extra actions for your custom source.
+Before create your new kind, you should have a look at existing kinds in
+rplugin/python3/denite/kind/*.py
+
+Instead of create new kind from scratch, you can extend one of the existing
+Kind class, let's take `openable` for example: >
```
lambdalisue: Vim's help does not use back-quote. I think it should be replaced with a double quote instead.

doc/denite.txt
```diff
@@ -1451,12 +1451,64 @@ abbr		(String)			(Optional)
...
+					*denite-kind-attribute-redraw_actions*
+redraw_actions (List) 				(Optional)
+		List of action names which redraw Denite buffer after
+		executed.
+
+					*denite-kind-attribute-action_*
+action_{name} (Function) 			(Optional)
+		Custom action logic with {name} as action name. The function
+		is passed with {self} and {context}. To get the seleted
+		candidates, you could use: `context['targets']`
```
lambdalisue: Same here

doc/denite.txt
```diff
@@ -1451,12 +1451,64 @@ abbr		(String)			(Optional)
...
+			target = context['targets'][0]
+			path = target['action__path']
+			" assume you have a MyEdit command
+			self.vim.call('MyEdit', path)
+>
+After create new Kind class, alias it to your source by >
+
+	self.kind = Kind(vim)
+<
+in the {__init__} method of your source class.
```
lambdalisue: Vim's help already use `{xxxxx}` syntax for an argument so it should not be used.

doc/denite.txt
lambdalisue: I noticed that an example code in `:h python` use a tab char..... I really don't like to use tab char for python that but if you prefer to use tab, I'm ok now.

chemzqm: Got it, I will use copy&paste to get avoid of this problem next time.

lambdalisue:
There are some left over for `{xxxxx}`

chemzqm:
@lambdalisue  
```
		Custom action logic with {name} as action name. The function
		is passed with {self} and {context}. To get the seleted
```
the usage of `{xxxxx}` here is the same as the ones of `:h denite-filter-attribute-filter`

lambdalisue:
Hum... OK.

Shougo:
Nice work!
Merged.

PR #190, opened 2017-02-22
Add keepalt for edit denite buffer

### LINKED ISSUE #189: Alternative buffer would be broken after invoke persist action
**Warning:  I will close the bug issue without the minimal init.vim and the reproduce ways.**

# Problems summary

An empty buffer named `[Denite]` would replace alternative buffer after invoke persist action

## Expected

Alternative buffer should not be replaced

## Environment Information (Required!)

 * denite version(SHA1): latest master

 * OS: MacOS

 * Vim/neovim version: NVIM v0.2.0-592-gf64f90b5 

 * `:CheckHealth` result(neovim only): all success

## Provide a minimal init.vim with less than 50 lines (Required!)

```vim
" Your minimal init.vim
set runtimepath+=~/path/to/denite.nvim/
```

## The reproduce ways from neovim starting (Required!)

 1. Open fileA with neovim then `:e fileB`
 2. Run `Denite file_rec`
 3. Select a candidate and choose action `preview` or `highlight`
 4. Close denite buffer by `esc`
 5. Hit `<C-^>` then the empty buffer would shown, but it should be `fileA`

chemzqm:
This would fix #189

### DIFF
--- a/rplugin/python3/denite/ui/default.py
+++ b/rplugin/python3/denite/ui/default.py
@@ -141,7 +141,8 @@ class Default(object):
             # Create new buffer
             self._vim.call(
                 'denite#util#execute_path',
-                'silent ' + self._context['direction'] + ' new', '[denite]')
+                'silent keepalt %s new ' % self._context['direction'],
+                '[denite]')
         self.resize_buffer()
         self._vim.command('nnoremap <silent><buffer> <CR> ' +
                           ':<C-u>Denite -resume -buffer_name=' +


rplugin/python3/denite/ui/default.py
```diff
@@ -141,7 +141,7 @@ def init_buffer(self):
             # Create new buffer
             self._vim.call(
                 'denite#util#execute_path',
-                'silent ' + self._context['direction'] + ' new', '[denite]')
+                'silent keepalt ' + self._context['direction'] + ' new', '[denite]')
```
lambdalisue: A lint say the line is too long. The line should be

```python
    'silent keepalt %s new' % self._context['direction'],
    '[denite]'
)
```

Or something similar I guess.

By the way, I agree with your change while denite is not a file-like buffer :+1:

chemzqm: @lambdalisue I've fixed that,  I counldn't open the travis log page due to bad network.

lambdalisue:
I leave the "merge" button in case Shougo wants some review.

Shougo:
Merged.

PR #195, opened 2017-02-25
Improve highlight of line source

chemzqm:
I need to make a new Source which inherit the line source, the old way of syntax definition can't works for inherited source.


Shougo:
OK

PR #202, opened 2017-02-27
Update neovim prompt

lambdalisue:
This PR does

- Update neovim-prompt
  - neovim-prompt no longer execute `redraw!` automatically
- For instance, denite.nvim execute `redraw!` just after `prompt.start()`

You can move `redraw!` after this PR wherever you want @Shougo


Shougo:
Merged.  Thank you.

PR #204, opened 2017-02-28
Bind scroll_page to previewwindow when possible

### LINKED ISSUE #191: Feature request: key mappings for scroll up and down the preview window
I want to navigate throught the preview window after preview action, but Denite prevent me from change current window, so the key mappings would be helpful.

chemzqm:
Attempt to fix #191.

Check previewwindow and bind scroll_page_backwards & scroll_page_forwards to previewwindow if  previewwindow exists.

This would make preview window more useful since denite prevent user change current window after open preview buffer.

### DIFF
--- a/rplugin/python3/denite/ui/default.py
+++ b/rplugin/python3/denite/ui/default.py
@@ -568,11 +568,34 @@ class Default(object):
     def scroll_window_downwards(self):
         self.scroll_down(self._scroll)
 
+    def __get_preview_window(self):
+        return next(filterfalse(lambda x:
+                                not x.options['previewwindow'],
+                                self._vim.windows), None)
+
     def scroll_page_backwards(self):
-        self.scroll_up(self._winheight - 1)
+        previewwindow = self.__get_preview_window()
+        if previewwindow:
+            prev_id = self._vim.call('win_getid')
+            wnr = previewwindow.number
+            wid = self._vim.call('win_getid', wnr)
+            self._vim.call('win_gotoid', wid)
+            self._vim.command(r'execute "normal! \<C-b>"')
+            self._vim.call('win_gotoid', prev_id)
+        else:
+            self.scroll_up(self._winheight - 1)
 
     def scroll_page_forwards(self):
-        self.scroll_down(self._winheight - 1)
+        previewwindow = self.__get_preview_window()
+        if previewwindow:
+            prev_id = self._vim.call('win_getid')
+            wnr = previewwindow.number
+            wid = self._vim.call('win_getid', wnr)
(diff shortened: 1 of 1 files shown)


Shougo:
Please add the documentation.

lambdalisue:
I think adding a new action for that is better while

1. this PR changes the current implementation and people would get confused if they just want to scroll the denite candidates
2. I'm not sure but is there any way to close a previewwindow from denite? If no, users have to quit denite to keep continue and it would be annoying

chemzqm:
@lambdalisue 

- It's not perfect solution, I think there would be idea(s)
- When previewindow is open by preview action the user most likely want to scroll the previewwindow (I think), maybe a variable which mark the previewwindow is opened by denite could be useful.
- The user would do preview action again to close the previwwinow opened by denite, at least that's how the preview action of file kind did, so if user have keymapping for preview, that just means type the key again.

lambdalisue:
> When previewindow is open by preview action the user most likely want to scroll the previewwindow (I think), maybe a variable which mark the previewwindow is opened by denite could be useful.

No. I meant if you add a new action, users can choose but if you overwrite the existing, there is no hope to use a previous behaviour. So I suggested to add a new action rather than changing a current behaviour (I'm ok if you change the default mapping to the new one if @Shougo allow)

> The user would do preview action again to close the previwwinow opened by denite, at least that's how the preview action of file kind did, so if user have keymapping for preview, that just means type the key again.

Make sense thanks 👍

Shougo:
Hm.
I will test the behavior later.

Shougo:
I have tested it.
Not so bad.

Please add the documentation for the behavior.

Shougo:
No progress?

Shougo:
I will implement `<C-w>P` behavior.
Closing.

petobens:
How can this be used to scroll the preview down without leaving the denite buffer? I tried adding the following action
```viml
function! s:scroll_preview_down(context)
    call denite#do_action(a:context, 'wincmd:P', a:context['targets'])
    execute 'normal! j'
    wincmd p
endfunction
call denite#custom#action('buffer,directory,file,openable,dirmark',
            \ 'scroll_preview_down', function('s:scroll_preview_down'))
```
which doesn't work.

Shougo:
You should learn Vim script and denite.nvim.

```vim
function! s:scroll_preview_down(context)
    wincmd P
    execute 'normal! j'
endfunction
call denite#custom#action('buffer,directory,file,openable,dirmark',
            \ 'scroll_preview_down', function('s:scroll_preview_down'), {'is_quit': 0})
```

petobens:
Thanks

Issue #207, opened 2017-03-03
`directory_rec` doesn't work on Windows


bennyyip:
This problem is caused by function [abspath](https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/util.py#L126)

```python
def abspath(vim, path):
    path = expand(path)
    return path if os.path.isabs(path) else os.path.join(
        vim.call('getcwd'), path)
```

`vim.call('getcwd')` return strings like "C:\Users\yeben\bennyyip.github.io" on Windows, which is not suitable for `os.path.join`. 
It should be "C:\\Users\\yeben\\bennyyip.github.io" or "C:/Users/yeben/bennyyip.github.io".
So I add a hacky line `    path.replace('\\', '/')` to solve this problem temporarily.

PR #208, opened 2017-03-03
fix directory_rec bug on windows

bennyyip:
monkey patch for #207

### DIFF
--- a/rplugin/python3/denite/util.py
+++ b/rplugin/python3/denite/util.py
@@ -125,6 +125,7 @@ def expand(path):
 
 def abspath(vim, path):
     path = expand(path)
+    path.replace('\\', '/')
     return path if os.path.isabs(path) else os.path.join(
         vim.call('getcwd'), path)
 


lambdalisue:
I'm not sure but I don't think `vim.call('getcwd')` is required here. Why don't you use the following (is this solve the problem on Windows? or introduce some bugs?)

```python
def abspath(vim, path):
     return os.path.abspath(expand(path))
```

bennyyip:
@lambdalisue I discovered this bug because file_rec  doesn't work on windows. I think my solution is buggy but it makes file_rec work again. That's why I open this issue, I hope @Shougo come up with an elegant solution. 
@Shougo changed this function from your version in a recent commit. Everything worked well  for me before, but I think he must has his reason for changing it.

lambdalisue:
> @Shougo changed this function from your version in a recent commit. Everything worked well for me before, but I think he must has his reason for changing it.

I see thanks. Well let's wait what Shougo think about this.

Shougo:
> I'm not sure but I don't think vim.call('getcwd') is required here. Why don't you use the following (is this solve the problem on Windows? or introduce some bugs?)

It does not work.
Because python cwd is different from Vim cwd.

Shougo:
Merged.

lambdalisue:
Hum... Then the following is better

```python
from os.path import normpath, join

def abspath(vim, path):
    return normpath(join(vim.call('getcwd'), path))
```

While the code above is equivalent to the `abspath` 
https://docs.python.org/3.6/library/os.path.html#os.path.abspath

Shougo:
OK.  I have improved it.

lambdalisue:
Oh... Yeah it should work but I would liked to wait a response from Windows user. Anyway, hope it solve the problem on Windows as well.

petobens:
After this changes the following command no longer works on Mac:
```
:Denite file_rec:~/git-repos
```
Instead I need to write it like:
```
:Denite file_rec:/Users/pedro/git-repos
```
Why isn't `~` expanded to my home directory? Thanks

bennyyip:
@petobens Actually, it's this commit break your command: https://github.com/Shougo/denite.nvim/commit/832e55abe32ad6619e3b6031481fee1d1cfe2a05

@Shougo forgot to expand path, I am going to pull a request to fix this.

chemzqm:
The bug is because
https://github.com/Shougo/denite.nvim/commit/832e55abe32ad6619e3b6031481fee1d1cfe2a05
removed the check for abspath.

petobens:
@bennyyip  @chemzqm thanks for the explanation.

janlazo:
Windows 8.1 user here.
Tested on recent Neovim build and master branch of denite.nvim.
Using `set shellslash` for file paths

`Denite directory_rec` outputs the following error message in `:messages`

[denite] Traceback (most recent call last):
[denite]   File "C:/Users/JanEdmund/AppData/Local/nvim/bundle/nvim/denite.git/rplugin/python3\denite\__init__.py", line 33, in start
[denite]     return self._uis[buffer_name].start(args[0], args[1])
[denite]   File "C:/Users/JanEdmund/AppData/Local/nvim/bundle/nvim/denite.git/rplugin/python3\denite\ui\default.py", line 103, in start
[denite]     self._denite.gather_candidates(self._context)
[denite]   File "C:/Users/JanEdmund/AppData/Local/nvim/bundle/nvim/denite.git/rplugin/python3\denite\denite.py", line 54, in gather_candidates
[denite]     candidates = source.gather_candidates(source.context)
[denite]   File "C:/Users/JanEdmund/AppData/Local/nvim/bundle/nvim/denite.git/.\rplugin/python3/denite\source\directory_rec.py", line 61, in gather_candidates
[denite]     context['__proc'] = Process(command, context, directory)
[denite]   File "C:/Users/JanEdmund/AppData/Local/nvim/bundle/nvim/denite.git/rplugin/python3\denite\process.py", line 24, in __init__
[denite]     cwd=cwd)
[denite]   File "C:\Python\35\lib\subprocess.py", line 676, in __init__
[denite]     restore_signals, start_new_session)
[denite]   File "C:\Python\35\lib\subprocess.py", line 955, in _execute_child
[denite]     startupinfo)
[denite] PermissionError: [WinError 5] Access is denied
[denite] Please execute :messages command.

Shougo:
@janlazo Please test the latest version of denite.

PR #212, opened 2017-03-06
Socket: Remove sleep and fix async progress

rafi:
Remove unnecessary `sleep` command and normalize `communicate` to correctly handle thread.


Shougo:
Nice.

PR #213, opened 2017-03-06
Source buffer highlight fix

rafi:
Regular-expression changes:
- 'Prefix' can contain another whitespace (otherwise, `%` or `#`)
- 'Modified' plus sign is _not_ wrapped with brackets


Shougo:
Merged.

PR #214, opened 2017-03-06
Fix buffer visibility detection ('a' or 'h' prefix)

rafi:
The function `bufwinnr` only searches current tab page.
Instead, we can use `win_findbuf` which was introduced in [7.4.1558](https://github.com/vim/vim/commit/9cdf86b86f5fdb5a45b682f336846f9d9a9c6f1f).

Also, re-indented the format operation for the sake of readability.


Shougo:
Merged.

PR #221, opened 2017-03-16
fix interactive grep

BonarBeavis:
Argument has to empty otherwise results are populated with files containing '!'.


Shougo:
Merged.

PR #229, opened 2017-03-29
Add command history source

pocari:
I have created command history source.
This source deal with `:history` command output in Denite interface.


Shougo:
Merged.

PR #238, opened 2017-04-08
Ignore winrestcmd when resuming a suspended denite

aeruder:
This fixes an easily recreated bug where the 'window' restore command is retrieved while the
denite window is open and upon finishing the denite action (and running this command) the remaining windows are "corrupted" (using far less rows than exist in the terminal).

How to recreate:

    :Denite file_rec
    " hit ctrl-z to suspend Denite
    " press enter to resume Denite
    " press enter again to open a file
    " screen is messed up.


Shougo:
Thanks.

PR #240, opened 2017-04-10
Make arguments required

chrisbra:
This avoids the error message `[denite] Empty sources` by making arguments required. New Error message is `E471: Argument required` which makes more clear that an additional argument is needed.

refereces vim-airline/vim-airline#1454


Shougo:
Thanks.

PR #241, opened 2017-04-10
Define path in context dictionary before using it

chrisbra:
Else this leads to an Error message, because the dictionary argument path is used before it is defined.


Shougo:
Fixed.

PR #246, opened 2017-04-13
Specify that '_' can be used to change the default for all sources

aeruder:
This can be used to change the default matchers, etc.


Shougo:
Merged.

PR #255, opened 2017-04-20
outline: fix some error on windows

zyujs:
1. `tempfile.NamedTemporaryFile(mode='w')` will create and lock the temp file, so ctags can't write to that file.
2. the default encoding of `open` is not utf-8, so an error may occur when opening a utf-8 encoding file that generated by ctags


Shougo:
Thanks

PR #266, opened 2017-04-28
scantree can now run with python 2 as well

MagicRedDeer:
The "file_rec" source uses scantree.py on windows in a separate python process. The python used for this purpose is the first 'python' found in environment path. If this python does not support the 'yield from' syntax, the process will return throwing 'SyntaxError' exception and the denite buffer will be empty. 

From the original code it seems like that scantree.py is meant to support more python versions, therefore any subsequent drop in performance, although minimal, should be acceptable.


Shougo:
Thanks

PR #267, opened 2017-05-03
Make ctags file option customizable

pinzolo:
[jstemmer/gotags: ctags\-compatible tag generator for Go](https://github.com/jstemmer/gotags) supports only `-f` option for writing to specified file.
I want to use `-f` option instead of `-o` option or customize this option like this pull request.

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -1216,7 +1216,7 @@ outline		Gather outline and jump to the target.
 				(default: ['ctags'])
 		options 	the default options for command
 				(default: [])
-		file_opt_name 	the default option name for writing to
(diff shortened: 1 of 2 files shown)


doc/denite.txt
```diff
@@ -1216,6 +1216,9 @@ outline		Gather outline and jump to the target.
 				(default: ['ctags'])
 		options 	the default options for command
 				(default: [])
+		file_opt_name 	the default option name for writing to
```
Shougo: It should be: `file_opt`.

Shougo:
Please change the option name.

Shougo:
Merged.  Thanks.

Issue #253, opened 2017-04-19
winrestcmd incorrectly restores windows when screen size has changed


aeruder:
Another one similar to #248 

To recreate:

    :Denite file_rec
    <resize terminal bigger>
    <select/accept file>

This series of actions creates a window that doesn't fill the entire nvim screen - have to use :resize to fix.

Shougo:
OK.  Infinite problems....
I gave up to restore.

languitar:
What is the future plan for this? I am currently in the situation that when a `cwindow` is open when I start denite, the `cwindow` will be as heigh as denite + cwindow after closing, which essentially wastes nearly the complete screen.

languitar:
Btw, I could also live with this if there was a `--no-split` option.

aeruder:
@languitar I think there is a better solution here.  I think anything using winrestcmd() is going to be fragile, but I'm wondering if we might be able to change the sequence of events to:

Starting denite:

* retrieve winrestcmd()
* open our denite window
* get current tab and window information from getwininfo()
* do denite things

When ready to close denite:

* get current tab and window information from getwininfo()
* If everything still matches we know that winrestcmd() is still valid
* close denite window
* run our saved winrestcmd()

Basically we just have to make sure that things don't change outside of denite's knowledge and if so, we can use winrestcmd() to handle issues like you are describing.  I'll see if I can put together a pull request today.

Shougo:
> Btw, I could also live with this if there was a --no-split option.

https://github.com/Shougo/denite.nvim/issues/217

Shougo:
@aeruder It seems the good idea.

aeruder:
Easier said than done, because there's also a whole bunch of issues surrounding bufnr if you do it this method (i.e. will bdelete! of the denite buffer close 0, 1, or more windows?) so I went back to a similar approach to starting, basically

Opening a denite window:
* save terminal/guiwindow size
* save getwininfo
* save winrestcmd
* Open denite

Closing a denite window:
* Close denite window
* Compare getwininfo() with saved - make sure the same windows are open in the tab
* Compare terminal/guiwindow size - needs to be the same
* Run winrestcmd if everything matches

Shougo:
Fixed.

languitar:
Sorry, but your new fix doesn't work in my case. #276 did work though.

Shougo:
@languitar Please create new issue for it.
I will test.

[...]

PR #276, opened 2017-05-15
Safely run winrestcmd when possible

aeruder:
This is another attempt at fixing #253 and
a bunch of similar issues.  We use getwininfo() to look
at the window state before running denite, and the window
state after running denite.  Anything looks bad? Just leave
things alone.


aeruder:
I'm testing this one for now, would appreciate some feedback from @languitar

languitar:
Yes, this looks much better. Seems to work well in my use cases.

Shougo:
I will create another PR...

PR #282, opened 2017-05-25
Remove double negation in sentence in docs

thalesmello:
A double negation makes the dentence a little bit harder to understand.
If we invert the question, the answer turns out to be clearer.


Shougo:
Thanks

PR #298, opened 2017-07-01
WIP: RFC: Support multiple rhs keys

lambdalisue:
Close #161.

Note that individual keys on rhs would not be resolved.

```
" The following does not work
call denite#custom#map('insert', '3<C-f>', '<Right><Right><Right>')

" The following should works
call denite#custom#map('insert', '3<C-f>', '<denite:move_caret_to_right><denite:move_caret_to_right><denite:move_caret_to_right>')
```


lambdalisue:
Oh. I noticed that this implementation prevent multiple lhs...

Shougo:
Oh....
😢

PR #299, opened 2017-07-03
Add ability to access/modify Denite context

wilywampa:
This implements the ability to read/write to a Denite instance's context dictionary. I included (in the documentation I wrote but also below) an example that shows how to implement a key that reverses the order of candidates using only `denite#context#get`, `denite#context#set`, and `denite#custom#map`. I hope that shows how useful this can be.

```vim
function! ToggleReverse() abort " {{{
    let sorters = split(denite#context#get('sorters'), ',')
    if index(sorters, 'sorter_reverse') == -1
        call add(sorters, 'sorter_reverse')
    else
        call remove(sorters, index(sorters, 'sorter_reverse'))
    endif
    call denite#context#set('sorters', join(sorters, ','))
    return '<denite:redraw>'
endfunction " }}}
call denite#custom#map(
    \ 'insert',
    \ '<C-s>',
    \ 'ToggleReverse()',
    \ 'noremap expr'
    \)
```

### DIFF
--- a/autoload/denite/context.vim
+++ b/autoload/denite/context.vim
@@ -1,0 +1,24 @@
+"=============================================================================
+" FILE: context.vim
+" AUTHOR: Shougo Matsushita <Shougo.Matsu at gmail.com>
+" License: MIT license
+"=============================================================================
+
+function! denite#context#get(...) abort
+  let key = a:0 ? a:1 : v:null
+  let bufname = a:0 > 1 ? a:2 : 'default'
+  if has('nvim')
+    return _denite_get_context(key, bufname)
+  else
+    return denite#vim#_get_context(key, bufname)
+  endif
+endfunction
+
+function! denite#context#set(key, value, ...) abort
+  let bufname = a:0 ? a:1 : 'default'
+  if has('nvim')
+    return _denite_set_context(a:key, a:value, bufname)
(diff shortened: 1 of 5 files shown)


wilywampa:
Working on the regression test failure.

Shougo:
I don't want to merge the feature.

* It is too long diff
* Non meaning Python code in context.vim

I will add the context change in the todo list.

Shougo:
I should add denite_change_context mapping instead.

Shougo:
@wilywampa For sorter_reverse, please create another PR. I will include it.

wilywampa:
@Shougo Done. But I'd like to know what's too complex about this implementation. I thought it was fairly elegant. I also don't understand what "Non meaning Python code in vim.vim" is trying to say.

wilywampa:
I'm not proud of the code in vim.vim, but I followed the example of what was already there to keep the function's variables in a temporary scope. I considered writing a simple class that would mean having less copy/pasted code but didn't think it was necessary for only 3 functions.

Shougo:
Your example is very hacky.
If the feature is useful, it should be supported by denite core.

Shougo:
get_context(), set_context() are hard to implement.  So your code is hacky

PR #316, opened 2017-07-16
Override quit/redraw in custom actions

wilywampa:
This is a very useful capability.


autoload/denite/custom.vim
```diff
@@ -107,15 +107,15 @@ function! denite#custom#action(kind, name, func, ...) abort
     if !has_key(custom, key)
       let custom[key] = {}
     endif
-    let custom[key][a:name] = a:func
+    let custom[key][a:name] = [a:func, a:0 ? a:1 : {}]
```
Shougo: OK.  I get it.

`a:func` should be `func` key member.

Shougo:
Merged.

PR #327, opened 2017-07-17
Improve/fix -cursor-pos

wilywampa:
This branch fixes that `-cursor-pos=+1` doesn't currently work (#323) and allows numbers other than 1 with the `+/-` notation.


Shougo:
The `-cursor-pos` document should be updated.

PR #329, opened 2017-07-17
Quit if -quit is on command line even if action is not is_quit

wilywampa:
This is useful for using `-immediately` for actions that don't quit denite.nvim. For example, my grep is not `is_quit`, so when I run `:Denite -resume -immediately -quit` it opens the denite window and it stays open despite having the `-quit` in the command line.

Basically, this allows non-`is_quit` actions to be selected using `-cursor-pos` but still keep denite closed after the action.


Shougo:
I think `-force-quit` option should be added instead the check of that `-quit` is added.
And please change the documentation.

Shougo:
Please fix the conflict.

Shougo:
Merged

PR #331, opened 2017-07-18
upgrade neovim-prompt

lambdalisue:
- Fix typing issues (mypy)
- Fix `mapleader` and `maplocalleader` issue (https://github.com/Shougo/denite.nvim/issues/330)
- Add `Prompt.harvest_interval` attribute to control harvest interval in subclass.

```python
class DenitePrompt(Prompt):
    def __init__(self, vim, context, denite):
        # ...

        # Regulate interval to 0.05
        self.harvest_interval = 0.05
```


Shougo:
Thank you!

PR #333, opened 2017-07-19
Fix document

pocari:
Fix following error.

```
[dein] Error generating helptags:
[dein] Vim(helptags):E154: Duplicate tag "unite-source-output" in file /Users/pocari/.cache/dein/.cache/init.vim/.dein/doc/unite.txt
[dein] function dein#recache_runtimepath[1]..dein#install#_recache_runtimepath[20]..<SNR>70_helptags, line 12
```


Shougo:
Thanks

PR #335, opened 2017-07-22
Make buffer source start faster

wilywampa:
I was noticing long times to start the buffer source when I had a couple hundred buffer (a lot, but not abnormal at work) so I decided to investigate. I found that the majority of the time it was taken to load was used computing the string length of the number of buffers. I don't know why it was slow, but I modified the source so it only does it once per gather instead of one per buffer.


Shougo:
I get it.
I have fixed it in better way.

PR #336, opened 2017-07-22
Add a feature to jump to next result in different file in grep source (and more!)

wilywampa:
I was missing being able to jump between files in grep results (very useful when there may be 50 results in one file) so I decided to add the feature. I made it general enough to be useful for other purposes, i.e. it can jump grouping by any attribute a candidate might have.

This example shows how to make `gj` and `gk` select between file groups instead of single candidates:

```vim
call denite#custom#map('normal', 'gj', '<denite:jump_to_next_by:path>')
call denite#custom#map('normal', 'gk', '<denite:jump_to_previous_by:path>')
```


Shougo:
OK.  It seems nice.

PR #341, opened 2017-07-25
Fix error on documentation

Leixb:
Documentation of `-winheight` talks about with when in fact modifies height as
it's name implies.


Shougo:
Thanks

PR #342, opened 2017-07-27
[RFC] implement highlighting for output:highlight

blankname:
It's currently a bit slow.
What would be the best way for me to cache the output of the highlight command (instead of calling it three times)?

If this sort of approach to highlighting is acceptable, I plan to implement highlighting for at least output:let and output:call.

### DIFF
--- a/rplugin/python3/denite/source/output.py
+++ b/rplugin/python3/denite/source/output.py
@@ -2,7 +2,7 @@
 # FILE: output.py
 # License: MIT license
 # ============================================================================
-
+import re
 import shlex
 import subprocess
 
@@ -16,6 +16,11 @@ class Source(Base):
         self.name = 'output'
         # why doesn't this seem to be working?
         self.default_action = 'yank'
+
+    def define_syntax(self):
+        cmd = self.context['args'][0]
+        if re.fullmatch(r'hi(ghlight)?(!)?', cmd):
+            self.define_syntax_for_highlight(cmd)
 
     def gather_candidates(self, context):
         args = context['args']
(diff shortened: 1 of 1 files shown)


rplugin/python3/denite/source/output.py
```diff
@@ -17,6 +17,55 @@ def __init__(self, vim):
...
+            ' deniteSource_outputHighlight_' + hi + 'xxx' +
+            ' ' + hi
+        )
+        self.vim.command(hi_name)
+        self.vim.command(hi_xxx)
+
+    def define_syntax(self):
+        cmd = self.context['args'][0]
+        if re.fullmatch(r'hi(ghlight)?(!)?', cmd):
+            self.vim.command('syntax include syntax/vim.vim')
```
Shougo: I think it should be in `self.deine_syntax_for_hi()`.
`define_syntax()` should not be long.

Shougo:
> What would be the best way for me to cache the output of the highlight command (instead of calling it three times)?

Nothing.  Unfortunately.
Vim/neovim should have the feature though...

blankname:
The major performance problem was caused by more highlights being defined each time `:Denite output:hi `was used.
Requiring two calls to `hi`, doesn't seem to have a large perf impact.
Performance seems acceptable now.

The current version seems to have introduced highlighting oddities to a few specific highlight entries:
image

Shougo:
Hm...
The highlight seems broken.
I think the previous version is better.

https://files.gitter.im/Shougo/denite.nvim/ijNm/thumb/denite-hi.jpg

Shougo:
The implementation is better.
But the highlight is worse...
Hm.  Pending now.

blankname:
The latest commit should fix the highlighting regression.

Remaining issues:
The `Msg` in `ErrorMsg` not being highlighted correctly.
Many highlight attributes not being highlighted correctly.

blankname:
I believe all major issues have been fixed:
image

Shougo:
Merged.

PR #345, opened 2017-07-28
add -immediately-1 option

tmn-o3:


Shougo:
Please fix the conflict.
Why this feature is needed?
Please explain it.

tmn-o3:
When I execute Denite tag for existing plural candidates, I'd like to choose a candidate on denite buffer.
But It has exactly one candidate, there is no need to open denite buffer.

Shougo:
OK.

hokorobi:
👍

PR #354, opened 2017-08-05
Detect validly non-windows env

delphinus:
because `context['is_windows']` may have `'0'` string that is truthy

I found this with MacVim 8.0.771 + Python3 3.6.2 (homebrew)

### DIFF
--- a/rplugin/python3/denite/filter/matcher_cpsm.py
+++ b/rplugin/python3/denite/filter/matcher_cpsm.py
@@ -28,7 +28,7 @@ class Filter(Base):
 
         if not self.__initialized:
             # cpsm installation check
-            ext = '.pyd' if context['is_windows'] else '.so'
+            ext = '.pyd' if context['is_windows'] != '0' else '.so'
             if globruntime(context['runtimepath'], 'bin/cpsm_py' + ext):
                 # Add path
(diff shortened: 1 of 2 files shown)


Shougo:
Thanks

rafi:
This change breaks macOS terminal support:
> [denite] matcher_cpsm: bin/cpsm_py.pyd is not found in your runtimepath.
> [denite] matcher_cpsm: You must install/build Python3 support enabled cpsm.

Expected behavior is loading `bin/cpsm_py.so`...

delphinus:
hmm, I also tested on macOS terminal.  In my env, `context['is_windows']` has `'0'` and it is truthy, but it has another value in other envs??

wilywampa:
I just used `matcher_cpsm` with no problem after this change on macOS Sierra. Are you sure you don't need to run `:UpdateRemotePlugins` and restart?

Shougo:
It is reproduced for me.
It must be fixed.

Shougo:
I will revert the change until the reproduce ways is uploaded.

Shougo:
I cannot reproduce your problem in Vim and neovim.

Shougo:
I should fix context initialization instead.

PR #358, opened 2017-08-07
Make the buffer source filterable

somini:
Just include the full file path for each buffer as `action__path`, if it exists. Otherwise, set it as `None`.

This is consistent with other sources and allows filtering in the buffer list.

My use case is restricting buffers to a project directory. So I use:
```vim
call denite#custom#alias('source', 'project/buffer', 'buffer')
call denite#custom#source('project/buffer', 'matchers',
			\ ['matcher_project_files', 'matcher_fuzzy'] )
```


Shougo:
Please fix Travis error.

PR #369, opened 2017-09-16
improve scantree.py by adding --ignore option

skeept:


doc/denite.txt
```diff
@@ -88,8 +88,6 @@ INSTALL							*denite-install*
 Note: denite requires Neovim(the latest is recommended) or Vim8.0+ with
 |:python3| enabled.
 
-Note: You need to install Python3.5+ or `pip3 install typing`.
```
Shougo: Please don't remove it.

Shougo:
Please fix the CI errors.

rplugin/python3/denite/source/file_rec.py
```diff
@@ -9,6 +9,29 @@
...
+def parse_command_for_scantree(cmd):
+    """Given the user choice for --ignore get the corresponding value"""
+
+    parser = argparse.ArgumentParser(description="parse scantree options")
+    parser.add_argument('--ignore', type=str, default='use_wildignore')
+
+    # the first name on the list is 'scantree.py'
+    args = parser.parse_args(cmd[1:])
+    if args.ignore == 'use_wildignore':
+        # how to get the value of vim variable?
```
skeept: @Shougo can you help here?
How to get the value of the option wildignore from vim.

I think this is possible but I don't know how to do it.

rplugin/python3/denite/source/file_rec.py
```diff
@@ -6,32 +6,30 @@
 
 from .base import Base
 from denite.process import Process
-from os import path, pardir
+from os import path
 from os.path import relpath, isabs, isdir, join
 from denite.util import parse_command, abspath
 import argparse
 import os
+import vim
```
skeept: This works for me when I run it from inside vim but it fails the CI checks.
Can you advice here how to fix this?

@Shougo

Shougo: Please don't use this.
Please use `self.vim` instead.

rplugin/python3/denite/source/file_rec.py
Shougo: `self.vim.options['wildignore']`

Shougo:
Merged.

skeept:
@Shougo  Thank you for merging. 

I was going to ask you if you had a chance of testing in your enviroment. I just use windows but I think it should work in other environments as well.

Shougo:
> I was going to ask you if you had a chance of testing in your enviroment. I just use windows but I think it should work in other environments as well.

I have tested it today.
The code is buggy.  I will fix it later.

Shougo:
I have fixed the errors.

skeept:
Thank you for the fixing the errors.. I looked at the code they are good.

Shougo:
:-)

Shougo:
> I was going to ask you if you had a chance of testing in your enviroment. I just use windows but I think it should work in other environments as well.

It works on Linux environment.

skeept:
It is necessary another small fix because python 3 doesn't always work in windows. Usually the python windows executable is python.exe. I added a small branch to fix that. Please review and merge when you have a chance.

Issue #376, opened 2017-10-05
File finding shows no matches after narrowing to a directory on Windows


spigo900:
# Problems summary
When you use `:Denite file` on Windows and browse into a directory, Denite shows no matches.

## Expected
`:Denite file` should show the files or directories in the directory you browsed to, if any exist.

## Environment Information (Required!)

 * denite version(SHA1):

 * OS: Windows

 * Vim/neovim version:

 * `:CheckHealth` result(neovim only):

```
health#nvim#check
========================================================================
## Configuration
  - SUCCESS: no issues found

## Performance
  - SUCCESS: Build type: RelWithDebInfo

## Remote Plugins
  - WARNING: "denite.nvim" is not registered.
  - WARNING: Out of date
    - SUGGESTIONS:
      - Run `:UpdateRemotePlugins`

health#provider#check
========================================================================
## Clipboard (optional)
  - SUCCESS: Clipboard tool found: win32yank

## Python 2 provider (optional)
  - INFO: Disabled. g:loaded_python_provider=1

## Python 3 provider (optional)
  - INFO: `g:python3_host_prog` is not set.  Searching for python in the environment.
  - INFO: Multiple python executables found.  Set `g:python3_host_prog` to avoid surprises.
  - INFO: Executable: C:\Users\crazy\AppData\Local\Programs\Python\Python35-32\python
  - INFO: Other python executable: /python
  - INFO: Python3 version: 3.5.2
  - INFO: python-neovim version: 0.1.13
  - WARNING: Could not contact PyPI to get latest version.
  - ERROR: HTTP request failed: error: curl error with https://pypi.python.org/pypi/neovim/json: 1

## Ruby provider (optional)
  - INFO: Ruby: ruby 2.3.1p112 (2016-04-26 revision 54768) [x64-mingw32]
  - WARNING: Missing "neovim" gem.
    - SUGGESTIONS:
      - Run in shell: gem install neovim
      - Is the gem bin directory in $PATH? Check `gem environment`.
      - If you are using rvm/rbenv/chruby, try "rehashing".
```

## Provide a minimal init.vim with less than 50 lines (Required!)

```vim
set runtimepath+=~/shared/nvim-data/plugged/denite.nvim/
```

## The reproduce ways from neovim starting (Required!)

 1. Use `:Denite file`.
 2. Browse into a directory.

## Screen shot (if possible)
[...]

PR #377, opened 2017-10-05
Fix file finding on Windows.

spigo900:
Fixes issue #376. The problem was that Denite uses '/' as its path 
separator on all platforms, which doesn't work (for some reason) on 
Windows.  This fixes by using `os.path.sep` instead of a literal 
forward-slash.


Shougo:
OK

PR #386, opened 2017-10-31
parse_tagline: return None when unparseable

aeruder:
I accidentally got emacs ctags in my path and it made a mess of some
tags files for perl causing denite to be unable to parse.

=== RIGHT NOW ===
parse_tagline returns {} when it fails:

    source/tag.py assumed that parse_tagline would be returning None when
    the parse failed.

    source/outline.py assumed it would always succeed and only went on
    to get KeyError's when the 'type' field wasn't filled in.

=== AFTER THIS CHANGE ===
Changing parse_tagline to return None on failure:

    source/tag.py already copes with it correctly

    source/outline.py updated to continue if None


Shougo:
Merged.

PR #388, opened 2017-11-02
Detect Haskell projects via stack.yaml file

dkasak:
Haskell projects commonly use Stack for building Haskell software. The presence of `stack.yaml` indicates a project root directory (see: https://docs.haskellstack.org/en/stable/yaml_configuration/).


Shougo:
OK, thanks

PR #396, opened 2017-11-28
Remove breaking empty argument in `rg` usage

ELLIOTTCABLE:
This suggested usage results in the following error:

    [denite] file_rec: : No such file or directory (os error 2)

/see #393


Shougo:
Merged.

Issue #300, opened 2017-07-03
Auto-resize (sometimes) not working with large screen


petobens:
I'm using:
```vim
call denite#custom#option('default', {
            \ 'auto_resize': 1,
            \ 'statusline': 0,
            \ 'winheight': 15,
            \ 'updatetime': 100,
            \ 'reversed': 1,
            \ 'prompt': '❯',
            \ 'prompt_highlight': 'Function',
            \ 'highlight_matched_char': 'Function',
            \ 'highlight_matched_range': 'Normal',
            \ 'highlight_mode_insert': 'WildMenu'
            \ })
```
and getting the following beaviour when running `:Denite file_mru`:

![jul-03-2017 20-06-38](https://user-images.githubusercontent.com/2583971/27809982-30cdc3e2-602b-11e7-8ed2-f7998b9fb50c.gif)

Shougo:
Please read the issue template.
The reproduce ways and the minimal vimrc are required!
I cannot reproduce the problem.

Shougo:
I will close the issue.  If you cannot create them.

Shougo:
And please describe the issue.
I cannot understand it.
We have not ESP skills.

Shougo:
@petobens I will close it.  If no response.

petobens:
I''ll try to find a way to reproduce it and report back. Please don't close yet.

Shougo:
OK.  I will wait.

petobens:
The problem seems to be with `ag`. With the following minimal vimrc:
```vim
set nocompatible

let $DOTVIM = expand('$HOME/.vim')

set runtimepath+=$DOTVIM/bundle/repos/github.com/Shougo/denite.nvim
filetype plugin indent on

if executable('ag')
	call denite#custom#var('file_rec', 'command',
        \ ['ag', '--follow', '--nocolor', '--nogroup', '-g', ''])
endif
```
If I run `Denite file_rec` then I can reproduce the behaviour as it can be seen in the following GIF:
![jul-22-2017 12-22-04](https://user-images.githubusercontent.com/2583971/28492365-68c62db4-6ed8-11e7-8881-5a9e0834e703.gif)

This also happens with the grep source when using `ag`. i.e:
```vim
if executable('ag')
	call denite#custom#var('grep', 'command', ['ag'])
	call denite#custom#var('grep', 'default_opts',
        \ ['--smart-case', '--vimgrep'])
[...]

Issue #172, opened 2017-02-04
Inconsisten reversed prompt behaviour


petobens:
Consider the following minimal vimrc:
```vim
set nocompatible

let $DOTVIM = expand('$HOME/.vim')

set runtimepath+=$DOTVIM/bundle/repos/github.com/Shougo/denite.nvim
filetype plugin indent on

call denite#custom#option('default', 'reversed', 1)
```

Start neovim with this minimal vimrc and then do  the following:
1) run `Denite file_rec`: you will see that the cursorline is on the very top of the denite buffer
2) start typing something to filter: the cursorline is now on the very bottom.

Withe `reversed` option I expect the cursorline to always start on the bottom candidate.

![feb-04-2017 13-42-31](https://cloud.githubusercontent.com/assets/2583971/22619979/d5243906-eadf-11e6-9b20-407e52152dc1.gif)

Shougo:
I will check it.

Shougo:
It is bug.  But it takes some time to fix.  It is not easy.

petobens:
Ok. Thank you.

petobens:
Thank you!

petobens:
@Shougo this problem has reappeared. You can reproduce it with my original instructions. Do you want me to open a new issue?

Shougo:
I will test it.

Shougo:
@petobens Please git bisect the problem.

petobens:
Were you able to reproduce the problem? I tried git bisecting but with several commit I cannot run `Denite file_rec` because  I get errors about missing`winwidth` or `cursor_shapes` keys.

Shougo:
Fixed.

petobens:
Thank you!

petobens:
@Shougo after your fix I can't move inside the buffer between candidates: the cursor is always stuck on the last line...

petobens:
Using `reversed: 0` fixes it (but then the reversed option is not set)

petobens:
Thank you

PR #399, opened 2017-12-13
Do "normal! zb" only when window is resized.

oaue:
The command "normal! zb" is only necessary when the buffer is actually resized.

### DIFF
--- a/rplugin/python3/denite/ui/default.py
+++ b/rplugin/python3/denite/ui/default.py
@@ -403,9 +403,6 @@ class Default(object):
         self._vim.current.buffer[:] = self._displayed_texts
         self.resize_buffer()
 
-        if self._context['reversed']:
-            self._vim.command('normal! zb')
-
         self.move_cursor()
 
     def update_status(self):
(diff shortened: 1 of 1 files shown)


Shougo:
Why the change is needed?
I don't understand.
Please upload the example.

Shougo:
It is pending until understanding.

Shougo:
And please test `Denite file_rec -reversed` in the repository.
I think it does not work.

Related issue:
https://github.com/Shougo/denite.nvim/issues/172

Shougo:
Hm...
I should just remove `normal! zb`.

oaue:
With vim 8, `Denite file_rec -reversed`, a large directory structure and a SSD drive, I experienced key losts when typing fast.
For example, when I type `abcdef` only `abdf` appears.

I found that was caused by `normal! zb` executed at each call to `resize_buffer`.

But if I simply remove it and have `hidden` set, I can reproduce #300: the second time the window is open, there is blank lines at the bottom.
open:
first
and then reopen:
second

With the `normal! zb` executed only when the buffer is effectively resized, I don't have key losts any more and I cannot reproduce #300.

My minimal vimrc:
```vim
set hidden  
let mapleader=" "
            
filetype off
set rtp+=~/.vim/bundle/denite.nvim/

map <Leader>df :Denite file_rec<CR>
        
call denite#custom#option('default', 'reversed', 1)
call denite#custom#option('default', 'auto_resize', 1)
filetype plugin indent on
syntax on
```

PR #406, opened 2018-01-03
Normalize file paths in unique algorithm

wilywampa:
This change checks if a `word` is a path, then compares on normalized path instead of the raw `word` so that e.g. only one of `~/_vimrc` and `/home/user/_vimrc` will show up.


Shougo:
OK

PR #409, opened 2018-01-07
Add option to hide empty lines in 'line' source.

uinput:
This patch adds an option to the `line` source to exclude empty lines.


Shougo:
OK.

PR #425, opened 2018-01-28
Support multiple select URI

ponko2:


Shougo:
It is already implemented.

PR #429, opened 2018-02-03
faster help processing

mhartington:
I had some updates locally that made processing help tags much faster. 
Thought it would be good to add them to upstream


Shougo:
Merged.

Issue #431, opened 2018-02-04
delete key behaves as left arrow key


ggerico:
**Warning:  I will close the issue without the minimal init.vim and the reproduction instructions.**
I uses denite for years now, first time it happens.
 
# Problems summary
the delete key has the behavior of the left arrow key

## Expected
the delete key should delete the last character of my search 

## Environment Information (Required!)

 * denite version(SHA1):
8e4aacacd5509795cb94c7e42790cc173baaa276

 * OS: macOs high sierra, iTerm2 3.1.5 (report terminal type: xterm-256color)

 * Vim/neovim version: NVIM v0.2.2

 * `:checkhealth` or `:CheckHealth` result(neovim only):

health#deoplete#check
========================================================================
## deoplete.nvim
  - OK: has("nvim") was successful
  - OK: has("python3") was successful
  - INFO: If you're still having problems, try the following commands:
    $ export NVIM_PYTHON_LOG_FILE=/tmp/log
    $ export NVIM_PYTHON_LOG_LEVEL=DEBUG
    $ nvim
    $ cat /tmp/log_{PID}
    and then create an issue on github

health#nvim#check
========================================================================
## Configuration
  - OK: no issues found

## Performance
  - OK: Build type: Release

## Remote Plugins
  - OK: Up to date

## terminal
  - INFO: key_backspace (kbs) terminfo entry: key_backspace=\177
  - INFO: key_dc (kdch1) terminfo entry: key_dc=\E[3~
  - INFO: $XTERM_VERSION=''
  - INFO: $VTE_VERSION=''
  - INFO: $SSH_TTY=''

health#provider#check
========================================================================
## Clipboard (optional)
  - OK: Clipboard tool found: pbcopy

## Python 2 provider (optional)
  - INFO: `g:python_host_prog` is not set.  Searching for python2 in the environment.
  - INFO: Executable: /usr/local/bin/python2
  - INFO: Python2 version: 2.7.14
  - INFO: python2-neovim version: 0.1.13 (outdated; from /usr/local/lib/python2.7/site-packages/neovim)
  - WARNING: Latest python2-neovim is NOT installed: 0.2.1

## Python 3 provider (optional)
  - INFO: `g:python3_host_prog` is not set.  Searching for python3 in the environment.
  - INFO: Executable: /usr/local/bin/python3
  - INFO: Python3 version: 3.6.4
  - INFO: python3-neovim version: 0.2.1
[...]

PR #432, opened 2018-02-04
Fixed delete key behaving as left key (#431)

JochemKlingeler:
After some hunting, and the tip of @petobens that this bug happend after the last two commits, I found this. Directly returning the action call seems to fix the issue of not being able to backspace in the search field.


Shougo:
Sorry, it is not correct fix.
It just reverting my change.

Shougo:
I will fix it later.
Please wait.

PR #437, opened 2018-02-07
Fix insufficient candidates on help source

notomo:
Denite help results is insufficient since #429 .  
Only the last `runtimepath/doc/tag`'s candidates are shown.  
I fixed it.


Shougo:
Merged.

PR #440, opened 2018-02-11
Allow namespace source, filter, kind

### LINKED ISSUE #436: Fix recursive source/filter load
https://github.com/Shougo/denite.nvim/commit/a17f37b18b94162311ec9fd4849656e9b3fd3986

is for enabling recursive source/filter load like `foo/bar.py` to make external plugin namespace.
However, the commit is not enough for example `foo/bar.py` and `hoge/bar.py` would be conflict while these would have the same name.

I'll fix it with some example source/filter

lambdalisue:
Close #436 

This PR allows plugin developers to use namespace style source, filter, or kind like

- `my_plugin`
- `my_plugin/featureA`
- `my_plugin/featureB`

It's important for preventing name confliction.

Additionally, I re-create several builtin sources, filters to use this new feature like `file/old`.
Previous style `file_old` also work.


Shougo:
OH, it is too huge changes.
I will test it later.

Shougo:
More testers are needed.

rplugin/python3/denite/denite.py
```diff
@@ -237,37 +239,58 @@ def load_sources(self, context):
...
                     self._filters[alias].path = path
 
     def load_kinds(self, context):
         # Load kinds from runtimepath
-        loaded_paths = [x.path for x in self._kinds.values()]
-        for path, name in find_rplugins(context, 'kind', loaded_paths):
-            module = importlib.machinery.SourceFileLoader(
-                'denite.kind.' + name, path).load_module()
-            kind = module.Kind(self._vim)
+        rplugins = import_rplugins('Filter', context, 'filter', [
```
lambdalisue: ぷぎゃー

Shougo:
@lambdalisue Ping.

lambdalisue:
> denite#helper#complete needs to be changed for completing sources have the namespace.

Thanks! I'll check it 👍

Shougo:
It seems no problem.
I will merge it in the weekend.

somini:
Since I saw the "help wanted" tag, I tested this on my [big Denite configuration](https://gitlab.com/somini/dotnvim/blob/master/init.vim#L811-943) and couldn't find any bugs either.

Many thanks for improving the command completion function.

Shougo:
Merged.

Issue #407, opened 2018-01-05
Feature request : Customize status line depending on args


JCavallo:
As far as I understand, the status line is managed [here](https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/ui/default.py#L409).

I would like to be able to force what is written in the middle (denite_statusline_sources) to an arbitrary value (buffer_name for instance).

Shougo:
https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/ui/default.py#L417 is correct.

> I would like to be able to force what is written in the middle (denite_statusline_sources) to an arbitrary value (buffer_name for instance).

Hm.  Interesting.

JCavallo:
The problem I see is how to handle different sources.

If there is only one source, it is possible to call a method for customization.

What is buffer_name used for right now ? I don't see it anywhere, maybe it is used without airline ?

Shougo:
> Feature request : Customize status line depending on args

It means the source args?
I don't understand it.

> What is buffer_name used for right now ?

It is `buffer_name` option?  I don't understand.

JCavallo:
I usually set ``buffer_name`` when I call a source through a mapping:

`Denite my_source:args -buffer-name=custom_string`

or programmatically in another source / plugin:

``denite#start([{'name': 'my_source', 'args': my_args}], {'mode': 'normal', 'buffer_name': 'custom_string'})``

I want to have ``denite_statusline_sources == 'custom_string'``. But as far as I understand it is not possible to customize it with the current denite.

Shougo:
Hm.  It is not supported.  Yes.

JCavallo:
Would you agree to using ``buffer_name`` for the status line if it is not ``'default'`` ?

Something like:

``Denite file_rec`` will have a ``file_rec (1/100)`` status line
``Denite file_rec -buffer-name='my_string'`` will be ``my_string (1/100)``

Shougo:
> Would you agree to using buffer_name for the status line if it is not 'default' ?

I think statusline plugin customization is better.

PR #442, opened 2018-02-12
Improve statusline object

somini:
This extends the statusline object `b:denite_statusline` with more
parameters (like `buffer_name`) and includes "raw" versions of existing
arguments.
The existing arguments are kept for backwards compatibility.

This should make it easire to configure custom statusline functions.

Related to #407


Shougo:
Merged.

PR #452, opened 2018-02-22
Pass column to quickfix list.

thisduck:
The quickfix action wasn't passing the column information to the quickfix list. I've added the column information.


Shougo:
Thanks

Issue #466, opened 2018-03-21
context overwrites final_opts in grep source


uinput:
Hi,

I'd like to specify the `final_opts` in my grep alias. However, I noticed that denite will just use my current folder path as the path passed to grep. Here is the piece of code I'm talking about:
```python
        if context['__paths']:
            args += context['__paths']
        else:
            args += self.vars['final_opts']
```
Is there any way to enforce a path I want to grep? In my case, the path would be `:/`, specifying the root path for a git project. Why was the `__paths` variable used here in the first place, if `.` is the default for `final_opts`?

Shougo:
> Is there any way to enforce a path I want to grep?

Nothing.  It is used for the default path.

uinput:
But I already set `final_opts` to my path, namely `:/`.
```
call denite#custom#var("grep/git", "final_opts", [":/"])
```
But this does not affect the source behavior. It will still look in the directory of my current file.
What am I doing wrong here? According to the documentation, `final_opts` is

> the final arguments (default: ["."] path to search argument)

Shougo:
> But this does not affect the source behavior. It will still look in the directory of my current file.
> What am I doing wrong here? According to the documentation, final_opts is

The documentation is wrong.
I will fix it later.

Shougo:
I should split `default_paths` and `final_opts`.

uinput:
What about this?
```python
        final_opts = self.vars['final_opts']
        if len(final_opts) == 1 and final_opts[0] == "":
            args += context['__paths']
        else:
            args += final_opts
```
If we pass `[""]` as final_opts, Denite will replace it with the contents of `__paths`. This won't break setups passing an empty list like `[]` but will still allow for custom paths.

Shougo:
It breaks default `['.']` behavior.

uinput:
So we could just set `final_opts = ['']` as default? One could set it to `['.']` manually then, if a relative path output is preferred.

uinput:
I referenced a pull request to propose this change.

Shougo:
No.  It is just complex change.
Adding "default_paths" feature is better instead.

Shougo:
I will fix the issue.  But it takes time.
Please wait.

Shougo:
```diff
diff --git a/doc/denite.txt b/doc/denite.txt
index df0e1d9..a36f898 100644
--- a/doc/denite.txt
+++ b/doc/denite.txt
[...]

PR #468, opened 2018-03-22
Prevent context from overwriting final_opts in grep source.

uinput:
Fixing #466.


Shougo:
It breaks current grep source behavior.
Because, "." is needed for grep instead of other grep tools.

PR #492, opened 2018-06-21
Fix ignore_globs separator on windows

y2kbugger:
Has been tested on both
Windows 7, python 3.6 and Android/Termux/linux python 3.6


Shougo:
Please fix the lint error.

PR #497, opened 2018-07-03
Fix the pattern string

ichizok:
https://docs.python.org/dev/whatsnew/3.7.html#re

Since Python 3.7, nested set emits FutureWarning.


Shougo:
Thanks

PR #523, opened 2018-08-24
Fix the broken :DeniteCursorWord

Mariappan:
In the grep optimisation checkin this snippet has been missed, due to that
DeniteCursorWord is expecting me to press enter before search which is
annoying :P

### DIFF
--- a/rplugin/python3/denite/source/grep.py
+++ b/rplugin/python3/denite/source/grep.py
@@ -210,6 +210,8 @@ class Source(Base):
             elif not isinstance(arg, list):
                 raise AttributeError(
                     '`args[2]` needs to be a `str` or `list`')
+        elif context['input']:
+            patterns = [context['input']]
         else:
             patterns = [util.input(self.vim, context,
                                    'Pattern: ', context['input'])]


Mariappan:
Oops just now seen, it has been removed intentionally.
Is it so ?

Shougo:
He said:

```
another question: Is there a chance to use the current word under the cursor as an initial pattern for the grep source, but without submitting the pattern such that I can edit it before actually triggering the source?
```

Shougo:
You should use grep arguments 2 instead.

```
						*denite-source-grep*
grep		Gather grep results and nominates them.
		The source executes [command, default_opts, recursive_opts,
		args[1]..., `pattern_opt, args[2]`..., separator, args[0]...
		or final_opts] command line.
		Note: The command result must be "path:linenr:text" pattern.

		Source args:
			0. path string or list of paths
			1. string or list of arguments
			2. pattern string or list of patterns
```

languitar:
Maybe flags to `DeniteCursorWord` could solve this for everyone. Something like the following flags could make sense to my mind:

* One flag to indicate whether the word under the cursor shall be used for the interactive pattern or the narrowing text (or both)
* One flag to indicate whether everything should be submitted directly, or whether the user has the chance to modify the pattern or not

Shougo:
OK.  I have reverted the commit.
https://github.com/Shougo/denite.nvim/commit/9102dc48e130710cb0a5be1cf0858f31b3c9dd9c

Shougo:
> Maybe flags to DeniteCursorWord could solve this for everyone. Something like the following flags could make sense to my mind:

I don't want to add the flag.  Because, the flag is not needed.

```
:Denite grep:::`input('', expand('<cword>'))`
```

Fixes your issue.

languitar:
Thanks @Shougo that works well.

PR #524, opened 2018-08-24
Added support to auto close the preview window based on user option

Mariappan:
When we use the auto-preview and quit the denite buffer by opening the file or pressing Ctrl-C, the preview window stays and does not close immediately.

Added an option autoclose-preview which will decide to close the preview window or not based on the option set.


Shougo:
No.  It is not needed.  Because the feature is already implemented.

https://github.com/Shougo/denite.nvim/blob/master/rplugin/python3/denite/ui/default.py#L592

preview window will be closed.

I think you have already opened before denite buffer open.

Shougo:
If the problem exists, please create the new issue.
I will check it later.

PR #528, opened 2018-09-10
fixed getting the latest command history

nyatsui:
In the command_history source, can not get the latest command history.


Shougo:
Merged.

PR #537, opened 2018-11-12
revert execute command of execute_path()

RyoSusami:
An error appeared in a907dc01 when executing Denite.

NVIM 0.3.2-861-g9f3fb5511 on Windows 10
Python 3.7.1, neovim-python 0.3.0

### DIFF
--- a/autoload/denite/util.vim
+++ b/autoload/denite/util.vim
@@ -35,7 +35,7 @@ function! denite#util#execute_path(command, path) abort
   endif
 
   try
-    execute a:command s:expand(a:path)
+    execute a:command '`=s:expand(a:path)`'
   catch /^Vim\%((\a\+)\)\=:E325/
     " Ignore swap file error
   catch


Shougo:
The revert should not be included.
Please read the long issue.

https://github.com/Shougo/defx.nvim/issues/32

```
`=`
```

ignores `wildignore` files.  *The feature cannot be disabled*.  You should not use the feature.
Note:  Oh, it is very ugly behavior

Shougo:
> An error appeared in a907dc0 when executing Denite.

I think the escape is needed.
Please provide the filename.

RyoSusami:
Although it is not finely tracked yet, it seems that an error appears when grep or file/rec when the number of files in the directory is roughly 2000 or more.

RyoSusami:
There is no error in places with more than 40,000 files. Sorry, I do not know the condition to reproduce the error.

RyoSusami:
There is no error in places with more than 40,000 files. Sorry, I do not know the condition to reproduce the error.

PR #538, opened 2018-11-16
initial colorscheme preview support

mhartington:
This adds the beginning support for preview in colorscheme source.

- Adds new Kind (colorscheme)
- changes colorschemes source to use the new kind

denite-color-preview-better


Shougo:
I will improve the implementation.

PR #560, opened 2019-01-14
Add matcher that expands ~ and env vars

delphinus:
When I'm using `file` / `file/old` sources, I often want to narrow candidates to find files under the typical directories such as `~` (`$HOME`), `$VIM`, `$GOPATH`, and so on.  It can be done with this matcher.  This expands `input_str` with `denite.util.expand()`, and converts as such below.

```
~/git/repo
  → /Users/username/git/repo
$VIMRUNTIME/syntax/python.vim
  → /usr/local/Cellar/neovim/0.3.3/share/nvim/runtime/syntax/python.vim
```

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -1785,6 +1785,11 @@ matcher/cpsm (matcher_cpsm)
 		Note: You must use Python3 support enabled cpsm. >
 			$ PY3=ON ./install.sh
 <
+						*denite-filter-matcher/expand*
+matcher/expand
+		A matcher which filters the candidates using texts expanded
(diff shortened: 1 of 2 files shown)


Shougo:
OK.  I get it.
I think to add another option is better.

delphinus:
nice work!  But I see it does not highlight candidates with all `input_str`'s.  Is it a spec?



This is a shot with `:Denite file/old -expand` and input `~/.cache init`.  I want highlight `/Users/jinnouchi.yasushi/.cache` in addition to `init`.

PR #570, opened 2019-02-05
WIP: convert to yarp

Shougo:
~~Note: It does not work in Vim8 yet.~~


Shougo:
I have restored old Vim8 compatibility module unfortunately.
nvim-yarp partially work in Vim8, but...
It is freezed when denite closes current buffer.

defx does not have the problem, so I think nvim-yarp and prompt compatibility problem.

PR #582, opened 2019-03-04
[WIP] Python36 is required

### LINKED ISSUE #529: Python 3.6.1 will be required in the future
In the future version, denite.nvim requires Python 3.6.1 like defx.nvim.

It will be in 2019.

Shougo:
Fix #529


Shougo:
Merged.

PR #586, opened 2019-03-14
Allow recursive file searching when python not on PATH

y2kbugger:
Add a backup for finding the python.exe when other methods fail. It may be possible to use the sys.executable method by default, but I don't know about cross platform compatibility.


Shougo:
Thanks.  But I think to use `sys.executable()` is better in default.
I will work for that.

PR #588, opened 2019-03-17
fix nvim_open_win args

pocari:
currently `nvim_open_win` arguments are `nvim_open_win({buffer}, {enter}, {config})`.

I fixed it.


Shougo:
Oh. If so, I need to fix other plugins.

PR #590, opened 2019-03-18
fix import in "directory_rec" by "file_rec" was deprecated.

lunarxlark:
Error message`[denite] directory_rec: Please use "file/rec" source insead.` is displayed when I use "directory_rec" source.
So, fixed import statement in "directory_rec" source. This PR are affected by commit id: 28ea739d1ada062d8933578600d485036154eec1.


Shougo:
Thanks

PR #599, opened 2019-03-23
Fix python executable for scantree

skeept:
In windows we cannot use sys.executable all the time.
In my case sys.executable is pointing to vim executable not for python executable.
So we need a more robust way of getting the python executable name.

It should be possible to directly use the function using python but I don't know how to get the result in that case.

The other fix is because the completion logic doesn't work in windows because the path separator is '\' instead of '/'.
I convert this before and then it works.

Please change what you find necessary to change.


Shougo:
> In windows we cannot use sys.executable all the time.
> In my case sys.executable is pointing to vim executable not for python executable.
> So we need a more robust way of getting the python executable name.

Too bad Windows.

PR #607, opened 2019-03-25
neovim spell source

tex:
neovim spell source (using vim.funcs)
also found that action_replace does not work correctly

### DIFF
--- a/doc/denite.txt
+++ b/doc/denite.txt
@@ -1677,6 +1677,15 @@ register	Gather the registers.
 						*denite-source-source*
 source		Gather Denite source names.
 
+						*denite-source-spell*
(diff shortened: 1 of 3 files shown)


Shougo:
I have fixed the implementation.

tex:
I am trying to learn how to create sources, can you describe why your fixes?

the 'call' instead of 'funcs' is because of vim compatibility?

what about the self._arg and self.context['__arg']? why not self.arg?

thanks!

Shougo:
> the 'call' instead of 'funcs' is because of vim compatibility?

Yes.

> what about the self._arg and self.context['__arg']? why not self.arg?

Using context is better.  Because, `self.arg` object is shared.

PR #609, opened 2019-03-29
[WIP] Split UI process

### LINKED ISSUE #509: Split UI process
Denite should split into two processes like deoplete.

* UI process

* Denite process

Shougo:
Fix #509


Shougo:
I have split the process.  But I cannot get the results.

Shougo:
I should remove prompt.nvim instead.

PR #656, opened 2019-06-12
Fix help tag to denite-key-mappings instead of denite-mappings

zchee:
Fix help tag to `denite-key-mappings` instead of `denite-mappings`.


Shougo:
Thanks

PR #660, opened 2019-06-13
Add sorter that does nothing (useful for cpsm)

tkonolige:
This sorter is useful for filters that apply ordering (cpsm in #659).


Shougo:
The feature is no means.
You can disables the sorter by `[]`.

PR #675, opened 2019-06-29
Improve command_history speed with lua

skt041959:
Hi Shougo,

I find the command_history source is too slow in my usage due to large number of histories.
The bottleneck should be lots of nvim api calls from Python client.

With my code, the speed has been improved a lot. Would you please take a review ?

Thanks

### DIFF
--- a/rplugin/python3/denite/source/command_history.py
+++ b/rplugin/python3/denite/source/command_history.py
@@ -21,6 +21,18 @@ class Source(Base):
         self.vars = {
             'ignore_command_regexp': []
         }
+        self.vim.exec_lua(
+            """
+            local function getHistory()
+                local histories = {}
+                local number = vim.api.nvim_call_function('histnr', {':'})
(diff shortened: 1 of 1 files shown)


tex:
Can lua be used to speed up MRU source too?

skt041959:
> Can lua be used to speed up MRU source too?

@tex Do you mean which MRU source? I didn't find any in current Denite.nvim.

tex:
Sorry, I meant Shougo's neomru.vim. Briefly looked at it and it heavily uses self.vim.call...

Shougo:
> Sorry, I meant Shougo's neomru.vim. Briefly looked at it and it heavily uses self.vim.call...

In neomru, formatting is slow.

I have improved it.

tex:
Thanks!!!

Shougo:
I have fixed the compatibility problem in Vim8

Issue #699, opened 2019-07-31
-quick-move parameter broken


dsirov:
**Warning:  I will close the issue without the minimal init.vim and the reproduction instructions.**

# Problems summary
-quick-move parameter does not work correctly - an error is shown (see below).

I think the problem was introduced in [https://github.com/Shougo/denite.nvim/commit/69fc365671bceb6ac240fb90b7790e2fa30df40d](url)

:Denite -quick-move file/rec

The following output is produced:
```
[denite] Traceback (most recent call last):
[denite]   File "/home/david/dev/vim/nvim/plugged/denite.nvim/rplugin/python3/denite/rplugin.py", line 25, in start                                                                                       
[denite]     return ui.start(args[0], context)
[denite]   File "/home/david/dev/vim/nvim/plugged/denite.nvim/rplugin/python3/denite/ui/default.py", line 63, in start
[denite]     self._start_sources_queue(context)
[denite]   File "/home/david/dev/vim/nvim/plugged/denite.nvim/rplugin/python3/denite/ui/default.py", line 150, in _start_sources_queue
[denite]     self._start(context['sources_queue'][0], context)
[denite]   File "/home/david/dev/vim/nvim/plugged/denite.nvim/rplugin/python3/denite/ui/default.py", line 208, in _start
[denite]     if self._context['quick_move'] and self.quick_move():
[denite] AttributeError: 'Default' object has no attribute 'quick_move'
[denite] Please execute :messages command.
```

## Expected
Quick move is displayed upon Denite ...

## Environment Information (Required!)

 * denite version (SHA1):
c6a52cc6d98db857fffa0a19a7c96bda0230adee
 * OS:
Fedora 29
 * Vim/neovim version:
:version
NVIM v0.3.6                                                                                                                                                                                               
Build type: RelWithDebInfo
Lua 5.3
Features: +acl +iconv -jemalloc +tui

 * `:checkhealth` or `:CheckHealth` result(neovim only):
```
health#denite#check
========================================================================
## denite.nvim
  - OK: has("python3") was successful
  - OK: Python3.6.1+ was successful
```

## Provide a minimal init.vim with less than 50 lines (Required!)
Plug 'Shougo/denite.nvim',              { 'do' : ':UpdateRemotePlugins' }

## How to reproduce problems from neovim startup (Required!)
:Denite -quick-move file/rec

## Generate a logfile if appropriate
:messages
```
[denite] Traceback (most recent call last):
[...]

PR #700, opened 2019-08-01
ui/default.py: quick naive FIX for broken -quick-move parameter (#699)

dsirov:
However, -quick-move=immediately still does not seem to work correctly


Shougo:
My fix is better and it fixes `-immedidately` error.

PR #702, opened 2019-08-02
[for Windows] Improve scantree.

Milly:
* Add `--type` option to scantree.py
  * Fix argument parsing in file/rec
  * Using scantree.py in directory_rec


Milly:
I tried using `nargs=argparse.REMAINDER` in `parse_command_for_scantree()` to pass all other arguments to `scantree.py`.
But it's buggy at current python library.
So parsing all arguments.

PR #722, opened 2019-11-17
Add denite#util#input

tamago324:
* Add `denite#until#input`
* Use `denite#until#input` instead of `util.input`.
* If press `<C-c>` while waiting for input like https://github.com/Shougo/defx.nvim/issues/174, Vim stops.


tamago324:
Thank you for merge!

PR #729, opened 2019-12-05
add check to v:version in plugin dir

kazukazuinaina:
Hello.

In this document, `denite.nvim` support Vim 8.0.0107 or above or Neovim 0.30.

Therefore, I added checking about `v:version < 800` in plugin dir.

Please check this Pull Request.

### DIFF
--- a/plugin/denite.vim
+++ b/plugin/denite.vim
@@ -4,7 +4,7 @@
 " License: MIT license
 "=============================================================================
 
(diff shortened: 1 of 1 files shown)


Shougo:
Thanks.
But I have added Vim8.0 check in init instead.

I don't want to add the check in plugin dir.

kazukazuinaina:
Ok, I got it. Thanks!

PR #733, opened 2019-12-16
Fix issue where Default has no attribute '_filter_floating' error happens sometimes

yasuoza:
Sometimes after `call dein#update()`, denite prints following error and fails to call denite functions.

```
Default has no attribute '_filter_floating'  
```

This may be rare case, but this patch solved the issue at that time.

P.S.
I tried to reproduce above error, but I couldn't.


Shougo:
Thanks.

PR #741, opened 2020-01-18
Don't close filter buffer when split floating

nomasprime:


autoload/denite/filter.vim
```diff
@@ -190,9 +190,12 @@ function! s:async_update() abort
 endfunction
 
 function! s:quit() abort
+  let context = g:denite#_filter_context
+
   if winnr('$') ==# 1
     buffer #
-  else
+  elseif (context['split'] !=# 'floating' &&
```
nomasprime: This should really call a function that returns boolean based on whether the split should be pinned to the top or not, i.e., taking `winrow` into account, but it's only my first day learning VimScript.

I'll try to update tomorrow.

Shougo: Fixed.

nomasprime: Thanks @Shougo

PR #759, opened 2020-03-06
add '-split=floating_relative' option

orokasan:
This PR adds an option for opening denite's floating window near cursor position like popup-menu.

output
`:Denite -split=floating_relative neoyank register`

Now this option doesn't work with -auto-resize option.
I will support this later.

### DIFF
--- a/autoload/denite/filter.vim
+++ b/autoload/denite/filter.vim
@@ -104,7 +104,15 @@ function! s:new_filter_buffer(context) abort
             \ 'width': str2nr(a:context['winwidth']),
             \ 'height': 1,
             \})
-    else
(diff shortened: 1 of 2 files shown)


Shougo:
> Now this option doesn't work with -auto-resize option.
> I will support this later.

OK.

tex:
Have you tried to open filter buffer when floating_relative window is open?
I did and I think position of the filter buffer is not correct. It is shifted one character to right.
And if the filter buffer is opened top of cursor position the filter buffer position is totaly wrong.

orokasan:
Sorry. I'll make PR for fix this problem.

tex:
Awesome! Thanks! This floating_relative is great idea and I like it lot!

PR #760, opened 2020-03-07
fix some issues with `-split=floating_relative` option.

orokasan:
ref #759 

I'm trying to fix the problem with -auto-resize option.
Please wait for the merge.

PR #763, opened 2020-03-08
fix -start-filter behaviour when use with -split=floating_relative

orokasan:
ref. #760 

Sorry for taking your time again.
This PR fix the problem and improve behavior of -split=floating_relative.


tex:
Tested this and it fixes the issue. Thank you!

PR #778, opened 2020-04-19
BugFix:Fix error in command_history with auto_action='preview' option

rickysaurav:
* Add preview to the persistant list for command_history to stop autoclosing of denit-buffer with auto_action='preview'


Shougo:
Thank you for the contribution.
But my fix is better.

PR #779, opened 2020-04-21
toggle_select work well with auto action

impactaky:
I found toggle_select doesn't work when `--auto-action='preview'` option is set.
 This PR changes the auto_action target from selected candidates to cursor line.


Shougo:
Merged.

PR #792, opened 2020-07-29
Fix outline missing pattern, and typo in language_force

skt041959:
1. fix typo of language_force
2. fix missing pattern in json mode
3. show scope instead of filename if possible in json mode


Shougo:
Merged.

PR #793, opened 2020-08-03
Update the path where cpsm module can be found.

gotgenes:
Recently the cpsm project merged the following pull request: https://github.com/nixprime/cpsm/pull/53

This removes the `cpsm_py.so` artifact from `bin/cpsm_py.so` and now it can only be found under `autoload/cpsm_py.so`.

This PR fixes the path where denite searches for the module.


Shougo:
Thank you!

gotgenes:
Thank you for denite!

PR #816, opened 2021-01-11
Fix ValueError raised by Path.relative_to

skt041959:


Shougo:
Your PR is broken.  `os` is not imported.

Shougo:
I have added the check for it.

skt041959:
OK, thanks.

PR #821, opened 2021-01-16
Add only_in_current_directory option to buffer source

youxkei:
When vim has the following state:

- current directory: `/tmp/dir`
- buffers
	- `No Name`
	- `/tmp/foo.txt`
	- `/tmp/dir-or-not.txt`
	- `/tmp/dir/bar.txt`
	- `/tmp/dir/foo/bar.txt`

the buffer source with `only_in_current_directory` enabled lists the following buffers:

- `/tmp/dir/bar.txt`
- `/tmp/dir/foo/bar.txt`

The buffer source with `only_in_current_directory` enabled is useful when you use it combined with file/rec as following:

```
call denite#custom#alias('source', 'buffer/cd', 'buffer')
call denite#custom#var('buffer/cd', 'only_in_current_directory', 1)

" combine buffer/cd with file/rec
:Denite buffer/cd file/rec
```

### DIFF
--- a/rplugin/python3/denite/source/buffer.py
+++ b/rplugin/python3/denite/source/buffer.py
@@ -6,6 +6,7 @@
 
 import typing
 
+from os import sep
 from pathlib import Path
 from time import localtime, strftime, time
 from sys import maxsize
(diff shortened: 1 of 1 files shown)


Shougo:
This option is not needed.
Because `matcher/project_files` already exists for it.

PR #825, opened 2021-01-29
fix path separator for file source 

orokasan:
File source doesn't work correctly on windows.
The cause is same as #819


Shougo:
Thanks

PR #829, opened 2021-02-02
Document more specific

ujihisa:
lambda function itself is supported by Vim8. The problem it focuses on here is not that but just about passing it into that specific function.


Shougo:
Thanks.

PR #830, opened 2021-02-04
fix preview_bat split direction

matsui54:
When executing `preview_bat` action, preview window appears under denite window not above it. This behaviour is different from `preview` action.

denite_preview


Shougo:
Thanks.

PR #833, opened 2021-02-07
improve preview_bat window appearance

matsui54:
Improved `preview_bat` window appearance with the following changes. This will be useful for grep preview.
- When `target['action_line']` is specified, jump to the line and highlight it.
- show only line numbers. Header file name seems to be not needed.


Shougo:
Thanks.

PR #836, opened 2021-02-16
Add preview action to help source.

matsui54:
Reused code from `kind/file.py` with some modifications.
Also, 'canidate' seemes to be typo for 'candidate', so I replaced.


Shougo:
Merged.

PR #837, opened 2021-02-17
Fix for help preview.

matsui54:
Thanks for merging and fixing #836.
Additional changes for preview action.
- Append '\V' to `action__pattern` to treat '*' and other special characters in tag literally.
- Do `normal! zz` to ensure that the cursor line is always centered. `cursor()` does not center line when the line to jump is already in the screen.


Shougo:
Merged.

PR #838, opened 2021-02-18
Improve behavior of closing preview_bat window.

matsui54:
Fixed these problems about closing preview bat_window.
- When `-auto-action=preview_bat`, preview window flickers while moving cursor horizontally and filtering.
- On Neovim, `close!` command deletes terminal buffer if `hidden` option is not set. This causes error when deleting previewed buffers because denite tries to delete non-existing buffers(E516).


Shougo:
Merged.

PR #841, opened 2021-02-25
Don't save preview_bat output.

matsui54:
Buffers that `preview_bat` action makes cannot be reused and remaining bat buffers increase RAM usage. Deleting preview_bat buffer each time will save RAM usage and make implimentation simpler. In paticular, this would be useful for `-auto-action=preview_bat`.


Shougo:
Merged.

PR #845, opened 2021-03-13
Add preview_terminal for preview using external command.

matsui54:
`preview_bat` shows the contents of files asynchronously and is useful when using with `auto_action`.
I think using terminal for previewing can be used in various ways as well as previewing files.
This PR enable users to easily create sources which have preview feature using external commands.

I wrote a simple example of `git log`([here](https://github.com/matsui54/dotfiles/blob/master/nvim/rplugin/python3/denite/source/git/log.py)).


Shougo:
OK.

PR #846, opened 2021-03-18
Open help in help window

matsui54:
I think it's better to use help window for opening help. In normal window, helptag jump can't be used.


Shougo:
OK

PR #847, opened 2021-03-22
Fix removing autocmd

matsui54:
## Problem
CursorMoved autocmd for auto-action isn't removed when opening the denite buffer of the same source. This slows denite when `auto-action` is specified.
## Cause
[This line](https://github.com/Shougo/denite.nvim/blob/8c44de41ec46c44f84dd70907a5763112df2eacb/rplugin/python3/denite/ui/default.py#L182) does not work because of these reasons.
- `autocmd! <buffer>` must be called with autocmd-event name.
- This command is called before entering denite buffer


Shougo:
Thanks.

PR #848, opened 2021-03-23
Add toggle-auto-action mapping

matsui54:
`auto_action` feature is useful, but sometimes previewing slow denite and it can be annoying.
It would be very useful if auto_action can be toggled without restarting denite.


Shougo:
Merged.

PR #853, opened 2021-05-01
Fix matcher/matchfuzzy not working

Bakudankun:
matcher/matchfuzzy throws errors on Vim8.
I think `'input'` is correct here, isn't it?


Bakudankun:
Tests failing because neovim nightly is temporary unavailable. neovim/neovim#14461

Shougo:
Thanks.

PR #863, opened 2021-11-01
expand wildcard in rtp

kols:
Wildcard is allowed in runtimepath. If not expanded properly, a rtp like `~/.vim/pack/*/start/*` will cause a failed search of the rplugin. And a 'Source xxx is not found' exception will be raised.


Shougo:
Merged.

